PR #24, opened 2015-02-05
Changed from ascii to utf8 in the string decoding, otherwise eg. Norwegi...

flikka:
...an characters makes the signature verification fail


brandwe:
:shipit: Thanks!

PR #30, opened 2015-04-27
return the callback for federationmetadata template on consecutive reque...

### LINKED ISSUE #22: identity SAML callback not called
There seems to be an issue with viewing a `/identity` route as in the example for SAML authentication. [examples/login-saml/app.js#L193](https://github.com/AzureAD/passport-azure-ad/blob/master/examples/login-saml/app.js#L193)

Where the first load of `/identity` works fine, but the next won't, as the callback is never called when it does not match the condition anymore `this.federationMetadata === null` see in [saml.js#L104-L111](https://github.com/AzureAD/passport-azure-ad/blob/master/lib/passport-azure-ad/saml.js#L104-L111)

Happy to submit a pull request with the following fix, if agreed to be an issue.

```javascript
if(this.federationMetadata === null) {
    templates.compile('federationmetadata.template.xml', params, function(err, data) {
      if(err === null) {
        self.federationMetadata = data;
      }
      callback(err, data);
    });
} else {
    callback(null, this.federationMetadata);
}
```

fpereira1:
...sts (fixes #22)

Fixes an issue that the callback is never called on a second render of the federation metadata because, it does not match the condition anymore `this.federationMetadata === null` - it would happen when you viewed the `/identity` route twice for example. The page would just never respond.

### DIFF
--- a/lib/passport-azure-ad/saml.js
+++ b/lib/passport-azure-ad/saml.js
@@ -108,6 +108,8 @@ SAML.prototype.identity = function (callback) {
       }
       callback(err, data);
     });
+  } else {
+   callback(null, this.federationMetadata);
   }
 };
 


msotclas:
Hi __@fpereira1__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
        This seems like a small (but important) contribution, so no Contribution License Agreement is required at this point. Real humans will now evaluate your PR.
    </span>

TTYL, MSOTBOT;

brandwe:
Taken. Thanks!

PR #48, opened 2015-08-27
Add ability for callers to specify the response type and the nonce parameter

SLesser:


SLesser:
This is not needed anymore

PR #58, opened 2015-09-02
Allow resource parameter, response mode, and response type options to be set

SLesser:
- Include the "resource" parameter when there is one...otherwise the request to get the access token will fail.

- Allow local options of response mode and response type to override the global options

### DIFF
--- a/lib/passport-azure-ad/oidcstrategy.js
+++ b/lib/passport-azure-ad/oidcstrategy.js
@@ -472,13 +472,14 @@ function(next) {
 
       var params = self.authorizationParams(options);
       var params = {};
-      params['response_type'] = self._responseType;
-      log.info('We are sending the response_type: ', self._responseType);
+      params['response_type'] = options.responseType || self._responseType;
+      log.info('We are sending the response_type: ', params['response_type']);
       params['client_id'] = config.clientID;
       params['redirect_uri'] = callbackURL;
-      params['response_mode'] = self._responseMode;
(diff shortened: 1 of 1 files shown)


brandwe:
Can you explain why you are using options instead of self_.*. Those are options that are loaded by the configuration method and already passed by the config.js. Did you look at the latest (1.3.0) when making the changes?

SLesser:
I was looking at 1.2 at that point...but isn't the config and self options set up during initialization rather than when making an authenticate call?  It seems that different post requests may want different responses (separate calls for id tokens versus access tokens).

Sent from Outlook<http://aka.ms/Ox5hz3>




On Tue, Sep 1, 2015 at 10:11 PM -0700, "Brandon Werner" <notifications@github.com<mailto:notifications@github.com>> wrote:


Can you explain why you are using options instead of self_.*. Those are options that are loaded by the configuration method and already passed by the config.js. Did you look at the latest (1.3.0) when making the changes?

-

brandwe:
Actually, I completely changed that for 1.3.0 for B2C. Now configuration is loaded at authentication time and then cached unless changed. This is because for B2C policies have their own metadata and can be changed at runtime. We don't know what policy we'll get. So, I just cache each policy metadata as it comes by.

https://github.com/AzureAD/passport-azure-ad/blob/master/lib/passport-azure-ad/oidcstrategy.js#L554

SLesser:
As discussed in email, this pull request is to enable the caller to specify different response types from their authenticate call while still using the same strategy object between the different calls.

PR #79, opened 2015-12-24
Fix strategy tests

ryanvog:
Several tests in the `oidc-tests.js` file were not actually exercising the product as they throw a ReferenceError, which was caught by the test.

This PR includes two fixes:
1. Fix `oidc-tests.js` tests to actually test the product under test.
2. Fix duplicate tokenURL validator configuration entry.
  a. The isURL validator indirectly checks for non-empty strings which was the other configured validator.


brandwe:
I'm ready to take this, but can you please remove the `// missing required checker` code comments. It could confuse others reading the code that it is actually missing when in reality you added it. :-)

brandwe:
I'm taking this now since I'm going to fix your other issues and want this in the code. I'll do the update I mention above. Thanks!

PR #94, opened 2016-04-07
Add Prompt to Config

bernays:
This allows for developers to require users to input credentials upon logging in everytime. Without this users must clear their cookies on login.microsoft.com in order to switch users on a web app using this strategy for authentication.

https://msdn.microsoft.com/en-us/library/azure/dn645542.aspx#code-snippet-3


brandwe:
Can you update the documentation with this switch as well? I'd also like this to go in to the /dev branch first.

PR #116, opened 2016-06-10
fix BearerStrategy.call _extend leaving out req. fixes #112

### LINKED ISSUE #112: done is not a function. bearerstrategy.js:149
I'm building a single page app to use AzureAD and i'm using adal-angular client side with passport-azure-ad server side

When I try to use the BearerStrategy it fails.

the BearerStrategy wraps the original passport-http-bearer and calls it internally like this
```
BearerStrategy.call(this, util._extend({}, opts, { passReqToCallback: true }), jwtVerify);
```

the jwtVerify function expects 3 arguments (req, token, done)

the req object doesn't get passed in and as a result the token and done function get shifted left. internally in jwtVerify req is the token and token is the done callback and done is undefined

this causes the ```jws.decode(token)``` to fail when it tries to decode the token because 'token' is a function.

If i change the jwtVerify function to remove req and just do (token, done) it still fails with TypeError: Invalid hex string
```
TypeError: Invalid hex string
    at TypeError (native)
    at Buffer.write (buffer.js:594:21)
    at fromString (buffer.js:113:26)
    at new Buffer (buffer.js:58:12)
    at Object.exports.getElement.exports.getFirstElement.exports.originalURL.exports.merge.exports.uid.exports.rsaPublicKeyPem
```
rsaPublicKeyPem uses new Buffer and that's where it fails when i ignore the missing req object for jwtVerify()

dhodgin:
Since opts is initialized at the top of the function to either an empty object or the options variable and the code afterwards assumes it will be an object and modifies its properties we can remove the need to extend the opts into an empty object in the BearerStrategy.call and just combine opts with passReqToCallback

util._extend does not accept more than 2 arguments so currently passReqToCallback is dropped and not forced true unless you set it in your config of your own app

see: https://github.com/defunctzombie/node-util/blob/master/util.js#L572

### DIFF
--- a/lib/bearerstrategy.js
+++ b/lib/bearerstrategy.js
@@ -185,7 +185,7 @@ function Strategy(options, verifyFn) {
 
   // force passReqToCallback to be `true` for our decoding verify wrapper
   /* eslint-disable no-underscore-dangle */
-  BearerStrategy.call(this, util._extend({}, opts, { passReqToCallback: true }), jwtVerify);
+  BearerStrategy.call(this, util._extend(opts, { passReqToCallback: true }), jwtVerify);
   /* eslint-enable no-underscore-dangle */
 
   this.name = 'oauth-bearer'; // Me, a name I call myself.


dhodgin:
I just realized i had made this PR against master not dev. Will you guys take care of this to include it in dev or do you want me to do something?

yarandoo:
Hi @dhodgin I am new to this repo but it seems your PR fixed the mentioned issue, though documentation seems now outdated and some conditional logic relying in "passReqToCallback" looks like not necessary anymore as this parameter is forced to be always true. Please see #141. I am happy to fix those.

PR #120, opened 2016-06-13
OIDC example bug fix and clean up

lovemaths:
Fixed the path issue. Rewrote the README file, and added the detailed
instructions to set up the config file and run the app.


examples/login-oidc/client_config_idtoken.js
```diff
@@ -0,0 +1,38 @@
...
+  exports.creds = {
+    // This is your app's 'REPLY URL' in AAD
+    returnURL: 'http://localhost:3000/auth/openid/return',
+
+    // replace <your_tenant_name> with your tenant name,
+    // which is something like: "*.onmicrosoft.com"
+    identityMetadata: 'https://login.microsoftonline.com/<your_tenant_name>/.well-known/openid-configuration',
+       
+    // This is your app's 'CLIENT ID' in AAD
+    clientID: '2abf3a52-7d86-460b-a1ef-77dc43de8aad',
```
polita: You probably want to remove this clientID.

examples/login-oidc/client_config_idtoken_code.js
```diff
@@ -0,0 +1,37 @@
...
+  exports.creds = {
+    // This is your app's 'REPLY URL' in AAD
+    returnURL: 'http://localhost:3000/auth/openid/return',
+
+    // replace <your_tenant_name> with your tenant name,
+    // which is something like: "*.onmicrosoft.com"
+    identityMetadata: 'https://login.microsoftonline.com/<your_tenant_name>/.well-known/openid-configuration',
+    
+    // This is your app's 'CLIENT ID' in AAD
+    clientID: '2abf3a52-7d86-460b-a1ef-77dc43de8aad',
```
polita: Remove?

examples/login-oidc/client_config_idtoken.js
```diff
@@ -0,0 +1,38 @@
+  // Don't commit this file to your public repos. This config is for first-run
+  //
+
+  exports.creds = {
+    // This is your app's 'REPLY URL' in AAD
+    returnURL: 'http://localhost:3000/auth/openid/return',
+
+    // replace <your_tenant_name> with your tenant name,
+    // which is something like: "*.onmicrosoft.com"
+    identityMetadata: 'https://login.microsoftonline.com/<your_tenant_name>/.well-known/openid-configuration',
```
dhodgin: it might be worth mentioning here in the comment that this can also be the tenant guid id as well. some examples use the guid and so it can be confusing which is valid when both are.

examples/login-oidc/client_config_idtoken.js
```diff
@@ -0,0 +1,38 @@
...
+       
+    // This is your app's 'CLIENT ID' in AAD
+    clientID: '2abf3a52-7d86-460b-a1ef-77dc43de8aad',
+    
+    // This is your app's 'key' in AAD. Required if you are doing 'id_token code' or 'code',
```
dhodgin: helps if you add `Required if, in the responseType, you are doing...` This has confused me in other examples in the past

examples/login-oidc/client_config_idtoken.js
```diff
@@ -0,0 +1,38 @@
...
+    
+    // we should have token passed back to us in a POST
+    responseMode: 'form_post',
+    
+    // additional scopes you may wish to pass
+    // scope: ['email', 'profile'], 
+    // if you have validation on, you cannot have users from multiple tenants sign in
+    validateIssuer: true,
+    
+    passReqToCallback: false,
```
dhodgin: this should have a comment explaining what it does. Especially since it modifies the arguments passed back to the passport verify callback so it's not even visible in the code what it refers to really.
how about something like...
`If this is true, the req object is passed back with the verify function for the passport strategy`

examples/login-oidc/client_config_idtoken.js
lovemaths: removed

examples/login-oidc/client_config_idtoken_code.js
lovemaths: removed

examples/login-oidc/client_config_idtoken.js
lovemaths: Yes, added this comment

examples/login-oidc/client_config_idtoken.js
lovemaths: added it

examples/login-oidc/client_config_idtoken.js
lovemaths: Yes that would be helpful. Added comments about the usage of passReqToCallback in the client_config_*.js and app.js.

polita:
:shipit:

PR #122, opened 2016-06-14
Added more comments to bearstrategy.js

lovemaths:
Added more comments on the options and usage of bearer strategy

### DIFF
--- a/lib/bearerstrategy.js
+++ b/lib/bearerstrategy.js
@@ -22,7 +22,9 @@
  */
 'use strict';
 
-const BearerStrategy = require('passport-http-bearer').Strategy;
+// This strategy inherts from the passport-http-bearer strategy
+const BearerStrategyBase = require('passport-http-bearer').Strategy;
+
 const util = require('util');
 const jwt = require('jsonwebtoken');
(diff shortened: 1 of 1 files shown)


lib/bearerstrategy.js
```diff
@@ -34,6 +36,13 @@ const jws = require('jws');
  * signature is:
  *
  *     function(token, done) { ... }
+ * or
+ *     function(req, token, done) { ... }
```
polita: Who sets up this pattern? The general pattern in JS is to add optional/additional arguments to the *end* of the arguments list so that there's no ambiguity. Can we change this now, or is it breaking?

lovemaths: This is the pattern of passport js, and req is not an "optional" argument. In order to use the signature with req, user has to explicitly set passReqToCallback true in the options variable when they do Passport.use(new strategy(options, verify)). Passport strategy will read the passReqToCallback value to decide if this verify callback's first parameter is req. I mentioned this between line 42-45 and 72.

brandwe:
:shipit:

polita:
:shipit:

PR #126, opened 2016-06-16
Refactor code in oidcstrategy.js

lovemaths:
Refactored the code which handles the profile value based on whether user
wants to load user info or not.

### DIFF
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -128,7 +128,10 @@ function Strategy(options, verify) {
   this._passReqToCallback = !!options.passReqToCallback;
 
   if (!options.identityMetadata) {
-    // default value should be https://login.microsoftonline.com/common/.well-known/openid-configuration
+    // For B2C, the value should be 
+    // https://login.microsoftonline.com/common/.well-known/openid-configuration
+    // Otherwise, the value should be the same url with the word 'common' replaced
+    // by your tenant name or id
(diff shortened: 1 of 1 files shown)


lib/oidcstrategy.js
```diff
@@ -342,33 +350,15 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
-                sub,
-                iss,
-                profile: (idToken) ? makeProfileObject(jwtClaims, jwtClaimsStr) : undefined,
-                jwtClaims,
-                accessToken,
-                refreshToken,
-                params,
-              });
+              
+              return onProfileLoaded(self, {req, sub, iss, profile, jwtClaims, accessToken, refreshToken, params});
```
polita: I'm confused about the original intention of this code. What should profile have been set to in this code path: the parsed body, or the jwtClaims? @brandwe, can you enlighten me?

benkroeger: if `_shouldLoadUserProfile` resolves `load === true`, this code tries to load a user profile from `userInfoURL` and `makeProfileObject` parses the response body of that call.

otherwise, `makeProfileObject` parses the previously retrieved `jwtClaims`

IMHO, this logic can be removed since the v2.0 endpoints don't provide a user info endpoint and thus, profile is always parsed from `jwtClaims`

PR #127, opened 2016-06-16
Solved issue 121: regex error in pem.js

lovemaths:
If there is a match, str.match(regex) always returns at least one group, where
the first group is always the entire match, and the rest are the capturing groups.
Therefore getFirstCapturingGroup returns an array of two groups, and the second
one is the content captured.

Also added some comments on top of pem.js.

### DIFF
--- a/lib/pem.js
+++ b/lib/pem.js
@@ -28,9 +28,24 @@
  * openssl req -x509 -nodes -days 365 -newkey rsa:2048 -sha1 -keyout private.pem -out public.pem
  */
 
+
+/*
+ * certificate string and private key string have the following format:
+ * 1. certificate string:
+ *    "-----BEGIN CERTIFICATE----- content ----END CERTIFICATE-----"
+ * 2. private key string:
+ *    "-----BEGIN PRIVATE KEY----- content -----END PRIVATE KEY-----"
+ *
+ * We define two regex (CERT_REGEX and PRIVATE_KEY_REGEX), in which the content 
+ * we are interested in is captured as a (and the only) capture group. We can 
+ * then use the getFirstCapturingGroup function below to get the content, and 
+ * use the removeRN function below to sanitize the content by removing the \r 
+ * and \n's in the content.
+ */
+
 /* eslint-disable max-len */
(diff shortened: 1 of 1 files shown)


lib/pem.js
```diff
@@ -28,20 +28,40 @@
...
+ * we are interested in is captured as a (and the only) capture group. We can 
+ * then use the getFirstCapturingGroup function below to get the content, and 
+ * use the removeRN function below to sanitize the content by removing the \r 
+ * and \n's in the content.
+ */
+
 /* eslint-disable max-len */
-const CERT_REGEX = /-+BEGIN\s+.*CERTIFICATE[^-]*-+(?:\s)+([a-zA-Z0-9+\/=\r\n]+)-+END\s+.*CERTIFICATE[^-]*-+/;
-const PRIVATE_KEY_REGEX = /-+BEGIN\s+.*PRIVATE\s+.*KEY[^-]*-+(?:\s)+([a-zA-Z0-9+\/=\r\n]+)-+END\s+.*PRIVATE\s+.*KEY[^-]*-+/;
+const CERT_REGEX = /-+BEGIN\s+.*CERTIFICATE[^-]*-+\s+([a-zA-Z0-9+\/=\r\n]+)-+END\s+.*CERTIFICATE[^-]*-+/;
```
polita: I recommend simplifying this regex to be stricter and easier to read. Suggest:
-+BEGIN\s+CERTIFICATE-+\s+([a-zA-Z0-9+\/=\r\n]+)\s+-+END\s+CERTIFICATE-+

lib/pem.js
```diff
@@ -28,20 +28,40 @@
...
+const PRIVATE_KEY_REGEX = /-+BEGIN\s+.*PRIVATE\s+.*KEY[^-]*-+\s+([a-zA-Z0-9+\/=\r\n]+)-+END\s+.*PRIVATE\s+.*KEY[^-]*-+/;
```
polita: recommend: 
-+BEGIN\s+PRIVATE\s+KEY-+\s+([a-zA-Z0-9+\/=\r\n]+)\s+-+END\s+PRIVATE\s+KEY-+

lib/pem.js
lovemaths: Changed.

lib/pem.js
lovemaths: Changed.

polita:
:shipit:

Issue #111, opened 2016-06-09
Who is responsible for OAuth 'state' persistence and verification?


silhouettes:
Looking through the code and dependencies, I can't find anywhere that the 'state' parameter is being persisted to the session, or verified after it is returned to us in the redirect URL.

Am I right in concluding that it is up to the consumer of this module to handle this? i.e. I need to do the following:

1. Generate state and persist to session
2. Pass the state to this module when the OAuth process starts
3. Retrieve the returned state from the params of the redirect URL
4. Verify the returned state against the state persisted in the session

If this is the case, it would be slightly different from what `passport-github` and `passport-windowslive` do, which is handle state persistence/verification for us by delegating it to the `passport-oauth2` module. If this is not the case, would you mind giving me a pointer to where this is happening, because I'm having trouble configuring it correctly?

silhouettes:
Hi! Just following up on this, in case it has been forgotten. In case it helps, this is a useful resource: https://tools.ietf.org/html/rfc6749#section-10.12.

This is a fairly important security feature, so any investigation you do would be extremely helpful!

lovemaths:
done

Issue #104, opened 2016-05-03
provide more detailed error message on OIDC failure


benkroeger:
Azure AD provides quite a lot of information when the OAuth dance fails. Several error properties are provided as `error_*` prefixed query parameter in the callback redirect.
Unfortunately, the OidcStrategy only provides the `error` parameter back to passport

lovemaths:
We capture err_description and put it in log, maybe need to capture err_uri ( #166 ), closing the issue.

PR #129, opened 2016-06-22
Added state check, err msg details, comments on strategy

lovemaths:
(Solved) Issue #111 
(Part-Fix) Issue #104 

1. added the state check
2. more details in the error message of internal oauth error
3. removed _cachekey field in strategy, and modified code accordingly
4. removed the duplicate assignment of _skipUserProfile
5. added comments on how to use `this`

### DIFF
--- a/lib/errors/internaloautherror.js
+++ b/lib/errors/internaloautherror.js
@@ -13,6 +13,14 @@ function InternalOAuthError(message, err) {
   Error.call(this);
   Error.captureStackTrace(this, InternalOAuthError);
   this.name = 'InternalOAuthError';
+  if (err && err.data) {
+    try {
+      let errlist = JSON.parse(err.data);
+      message += '\nerror: \n' + errlist['error'] + '\nerror_message: \n' + errlist['error_message']; 
+    } catch(ex) {
+      message += '\ndata: \n' + err.data;      
+    }
+  }
   this.message = message;
   this.oauthError = err;
 }
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -47,6 +47,10 @@ const Metadata = require('./metadata').Metadata;
 // global variable definitions
 const log = new Log('AzureAD: OIDC Passport Strategy');
 
+// save the state used in authentication request here, so we can compare it with
+// the returning state from the response
+var saved_state = null;
+
 const memoryCache = cacheManager.caching({ store: 'memory', max: 3600, ttl: 1800 /* seconds */ });
 const ttl = 1800; // 30 minutes cache
 // Note: callback is optional in set() and del().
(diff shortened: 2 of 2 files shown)


lib/errors/internaloautherror.js
```diff
@@ -13,6 +13,13 @@ function InternalOAuthError(message, err) {
   Error.call(this);
   Error.captureStackTrace(this, InternalOAuthError);
   this.name = 'InternalOAuthError';
+  if (err && err.data) {
+    try {
+      let errlist = JSON.parse(err.data);
+      message += '\nerror: \n' + errlist['error'] + '\nerror_message: \n' + errlist['error_message']; 
+    } catch(ex) {
```
polita: If you can't JSON parse the error, maybe just log the error data as the message directly so it's not lost.

polita:
Are there bugs for the issues you're fixing here? If so, please link to them. If not, please file them.

lib/errors/internaloautherror.js
lovemaths: Changed.

lovemaths:
@polita bug reference added

polita:
Thanks. Why only "part-fix" of #104? What's remaining?

lib/oidcstrategy.js
```diff
@@ -47,6 +47,10 @@ const Metadata = require('./metadata').Metadata;
 // global variable definitions
 const log = new Log('AzureAD: OIDC Passport Strategy');
 
+// save the state used in authentication request here, so we can compare it with
+// the returning state from the response
+var saved_state = null;
```
silhouettes: Correct me if I'm wrong, but using a global variable like this means this will break when you have more than one OAuth process running concurrently. I think it needs to be stored per user/session. This is one of the ways that passport-oauth2 handles it: https://github.com/jaredhanson/passport-oauth2/blob/master/lib/state/session.js

lib/oidcstrategy.js
```diff
@@ -506,13 +585,11 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
-          let state;
-
-          if (!options.state) { state = utils.uid(16); } else { state = options.state; }
-
-          params.state = state;
+          // Add an automatically generated state to params, and remember it.
+          // Later when we get the id_token or authorization code back, we can 
+          // verify the state we sent are the same as the state returned in the
+          // query or body.
+          saved_state = params.state = utils.uid(16);
```
silhouettes: Is `params` intended to be an input to this module, an output, or both? When I first looked at it, I had assumed that I could pass in the value of `state` when I called this module to 'override' the state generation with my own mechanism. Whereas this line implies that you're overwriting whatever value of 'state' is passed in.

I don't personally need this feature, but it does seem like a point of confusion.

lib/oidcstrategy.js
lovemaths: Yes you are right. Thanks, I will fix it.

PR #130, opened 2016-06-22
Added unit test for pem.js

lovemaths:
Added unit tests for getPrivateKey, getCertificate and certToPem functions.


polita:
It looks like the builds are failing. If you need help figuring out how the travis build works, ping @brandwe.

lovemaths:
@polita Figured out the problem, it turns out that fs.readFileSync requires absolute path in travis build. Now it works.

test/pem_test.js
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+var extractAndCompare = function(extractFunc, pemFileName, correctValue) {
+  var pem = null;
+  var pemKeyOrCert = null;
+
+  try {
+    pem = fs.readFileSync(__dirname + '/resource/' + pemFileName, 'utf8');
+    pemKeyOrCert = extractFunc(pem);
+  } catch (e) {
```
polita: What should we be doing here if there's an exception? Swallowing it silently probably isn't the right thing.

test/resource/correctKeyOrCert.js
```diff
@@ -0,0 +1,50 @@
+exports.privateKey =
```
lovemaths: I generated them using openssl with some fake values, so they have nothing sensitive.

test/pem_test.js
```diff
@@ -0,0 +1,101 @@
...
+  'test certToPem': (test) => {
+    test.expect(1);
+    // tests here
+
+    var pemContent = pem.certToPEM('myCertificate');
+
+    var pass = ('myCertificate' === pem.getCertificate(pemContent));
+
+    test.ok(pass, 'should get the same certificate');
+
```
lovemaths: removed.

test/pem_test.js
lovemaths: I just rewrote the test to print the error message in case of exception.

polita:
:shipit:

PR #132, opened 2016-06-25
Removed whitespace in AuthnRequest/Issuer

yesitspeter:
This whitespace, unfortunately gets transmitted.  If it's present (re: the leading and trailing newlines) the ADFS server I have set up actually can't recognize the right relying party trust.


lovemaths:
@yesitspeter Please submit a pull request against release2x branch for the saml.js change, thanks.

PR #134, opened 2016-06-28
(Solved) Issue #111

lovemaths:
Added state check and unit test, and comments on strategy

1. added the state check and unit test. In order to test state, we have to use mocha + chai + chai-passport-strategy, and all the old unit tests using nodeunit should switch to this combo in the future.
2. removed _cachekey field in strategy, and modified code accordingly
3. removed the duplicate assignment of _skipUserProfile
4. added comments on how to use `this`


polita:
Please have @brandwe look this over as well, since it's a pretty complex change.

PR #138, opened 2016-06-30
key id check condition fixed

lovemaths:
"!keys.kid===kid" is "(!keys.kid) === kid", which condition always yields false. Therefore this bug causes us to always use the first key even the kid doesn't match, and that leads to token validation error. It should be "keys.kid !== kid".

### DIFF
--- a/lib/metadata.js
+++ b/lib/metadata.js
@@ -183,7 +183,7 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
     log.info('working on key:', key);
 
     // are we working on the right key?
-    if (!key.kid === kid) {
+    if (key.kid !== kid) {
       return false;
     }
 


lib/metadata.js
```diff
@@ -183,7 +183,7 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
     log.info('working on key:', key);
 
     // are we working on the right key?
-    if (!key.kid === kid) {
+    if (key.kid !== kid) {
```
brentschmaltz: Without looking at all the code. Suppose the metadata has 3 keys, is there a code path find the 'right' key?
If the 'right' key is not found, there needs to be a way to inform the developer of this fact.

polita:
:shipit:

PR #139, opened 2016-07-01
key id check condition fixed

lovemaths:
"!keys.kid===kid" is "(!keys.kid) === kid", which condition always yields false. Therefore this bug causes us to always use the first key even the kid doesn't match, and that leads to token validation error. It should be "keys.kid !== kid".

The rsaPublicKeyPem in aadutils.js of dev branch also has problem, the `encodedPubKey` has some extra '\n' inside which yield wrong pem.

Unit tests are added.

### DIFF
--- a/lib/metadata.js
+++ b/lib/metadata.js
@@ -176,8 +176,12 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
   let pubKey = null;
   let foundKey = false;
 
-  if (!(kid && keys)) {
-    throw new Error('keys or kid is missing');
+  if (!kid) {
+    throw new Error('kid is missing');
+  }
+
+  if (!keys) {
+    throw new Error('keys is missing');
   }
 
   keys.some((key) => {
@@ -206,11 +210,13 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
 
     pubKey = aadutils.rsaPublicKeyPem(modulus, exponent);
     foundKey = true;
-    return true;
+    
+    return pubKey;
   });
 
   if (!foundKey)
     throw new Error(`a key with kid %s cannot be found`, kid);
+
   if (!pubKey)
     throw new Error(`generating public key pem failed for kid: %s`, kid);
 
--- a/lib/metadata.js
+++ b/lib/metadata.js
@@ -174,16 +174,17 @@ Metadata.prototype.updateOidcMetadata = function updateOidcMetadata(doc, next) {
 Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
   const keys = this && this.oidc && Array.isArray(this.oidc.keys) ? this.oidc.keys : null;
   let pubKey = null;
+  let foundKey = false;
 
   if (!(kid && keys)) {
-    return null;
+    throw new Error('keys or kid is missing');
   }
 
   keys.some((key) => {
     log.info('working on key:', key);
 
     // are we working on the right key?
-    if (!key.kid === kid) {
+    if (key.kid !== kid) {
       return false;
     }
 
@@ -204,8 +205,14 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
     const exponent = new Buffer(key.e, 'base64');
 
     pubKey = aadutils.rsaPublicKeyPem(modulus, exponent);
-    return pubKey;
+    foundKey = true;
+    return true;
   });
+
+  if (!foundKey)
+    throw new Error(`a key with kid %s cannot be found`, kid);
+  if (!pubKey)
+    throw new Error(`generating public key pem failed for kid: %s`, kid);
 
   return pubKey;
 };
--- a/test/metadata_test.js
+++ b/test/metadata_test.js
@@ -187,17 +187,21 @@ exports.metadata = {
     test.expect(5);
 
     const m = new Metadata('www.example.com', 'oidc', {});
+
     // example public keys from AAD
+
     var key1 = {
       e: "AQAB", 
       kid: "MnC_VZcATfM5pOYiJHMba9goEKY", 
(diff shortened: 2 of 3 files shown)


lib/metadata.js
```diff
@@ -174,16 +174,17 @@ Metadata.prototype.updateOidcMetadata = function updateOidcMetadata(doc, next) {
 Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
   const keys = this && this.oidc && Array.isArray(this.oidc.keys) ? this.oidc.keys : null;
   let pubKey = null;
+  let foundKey = false;
 
   if (!(kid && keys)) {
-    return null;
+    throw new Error('keys or kid is missing');
```
brentschmaltz: It would be helpful to have two different errors. Distinguish between 'kid' and 'keys'.

lib/metadata.js
```diff
@@ -204,9 +205,15 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
...
 
     pubKey = aadutils.rsaPublicKeyPem(modulus, exponent);
-    return pubKey;
+    foundKey = true;
+    return true;
   });
 
+  if (!foundKey)
+    throw new Error(`a key with kid %s cannot be found`, kid);
+  if (!pubKey)
```
brentschmaltz: Normally we seem to add a blank line between clauses.
This would mean adding a blank line before: if (!pubKey)

test/metadata_test.js
```diff
@@ -183,4 +183,68 @@ exports.metadata = {
...
+  'check keys and pem generation': (test) => {
+    test.expect(5);
+
+    const m = new Metadata('www.example.com', 'oidc', {});
+    // example public keys from AAD
+    var key1 = {
+      e: "AQAB", 
+      kid: "MnC_VZcATfM5pOYiJHMba9goEKY", 
+      n: "vIqz-4-ER_vNWLON9yv8hIYV737JQ6rCl6XfzOC628seYUPf0TaGk91CFxefhzh23V9Tkq-RtwN1Vs_z57hO82kkzL-cQHZX3bMJD-GEGOKXCEXURN7VMyZWMAuzQoW9vFb1k3cR1RW_EW_P-C8bb2dCGXhBYqPfHyimvz2WarXhntPSbM5XyS5v5yCw5T_Vuwqqsio3V8wooWGMpp61y12NhN8bNVDQAkDPNu2DT9DXB1g0CeFINp_KAS_qQ2Kq6TSvRHJqxRR68RezYtje9KAqwqx4jxlmVAQy0T3-T-IAbsk1wRtWDndhO6s1Os-dck5TzyZ_dNOhfXgelixLUQ"
+    };
```
brentschmaltz: add a blank line here

test/metadata_test.js
```diff
@@ -183,4 +183,68 @@ exports.metadata = {
...
+    var key2 = {
+      e: "AQAB",
+      kid: "YbRAQRYcE_motWVJKHrwLBbd_9s",
+      n: "vbcFrj193Gm6zeo5e2_y54Jx49sIgScv-2JO-n6NxNqQaKVnMkHcz-S1j2FfpFngotwGMzZIKVCY1SK8SKZMFfRTU3wvToZITwf3W1Qq6n-h-abqpyJTaqIcfhA0d6kEAM5NsQAKhfvw7fre1QicmU9LWVWUYAayLmiRX6o3tktJq6H58pUzTtx_D0Dprnx6z5sW-uiMipLXbrgYmOez7htokJVgDg8w-yDFCxZNo7KVueUkLkxhNjYGkGfnt18s7ZW036WoTmdaQmW4CChf_o4TLE5VyGpYWm7I_-nV95BBvwlzokVVKzveKf3l5UU3c6PkGy-BB3E_ChqFm6sPWw"
+    };
+    var correctPemKey2 = "-----BEGIN RSA PUBLIC KEY-----\n" +
```
brentschmaltz: where is 'correctPemKey2' used?

test/metadata_test.js
```diff
@@ -183,4 +183,68 @@ exports.metadata = {
       'Should not fail with url present and auth type oidc'
     );
   },
+  'check keys and pem generation': (test) => {
+    test.expect(5);
+
+    const m = new Metadata('www.example.com', 'oidc', {});
+    // example public keys from AAD
```
brentschmaltz: add a blank line before a comment

lib/metadata.js
```diff
@@ -204,9 +205,15 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
     const exponent = new Buffer(key.e, 'base64');
 
     pubKey = aadutils.rsaPublicKeyPem(modulus, exponent);
-    return pubKey;
+    foundKey = true;
+    return true;
```
brentschmaltz: here the return is 'true', below the return is 'pubkey'. Why is that OK?

lib/metadata.js
```diff
@@ -204,9 +209,17 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
     const exponent = new Buffer(key.e, 'base64');
```
lovemaths: keys.some( condition_function ) is equivalent to:

for (key in keys) {
  if condition_function returns true
    break;
  else
   continue;
}

here it's using pubKey as true or false. Yes I should return pubKey, already changed it in the new commit.

test/metadata_test.js
lovemaths: done

polita:
Please ping @brentschmaltz to get a :shipit: for this PR.

lib/metadata.js
```diff
@@ -204,9 +209,17 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
...
 
     pubKey = aadutils.rsaPublicKeyPem(modulus, exponent);
+    foundKey = true;
+    
     return pubKey;
```
brentschmaltz: 1. should we just return 'true'. Why return pubKey?
2. do we need the 'const modulus = new Buffer(key.n, 'base64'); and const exponent = new Buffer(key.e, 'base64'); 
or can we just pass them as parameters as in:
pubKey = aadutils.rsaPublicKeyPem(new Buffer(key.n, 'base64'), new Buffer(key.e, 'base64'));

brentschmaltz:
answer the question on the Buffer and then commit.

Issue #90, opened 2016-03-18
Cannot read property 'keys' of undefined


AlexandrTuniev:
Hey
We have a node js server, that running using your library for authentication.
There is a exception thrown from it.
```
[2016-03-18 16:58:08.180] [ERROR] errorHandler - Cannot read property 'keys' of undefined
TypeError: Cannot read property 'keys' of undefined
    at Metadata.generateOidcPEM (xxx\server\node_modules\passport-azure-ad\lib\passport-azure-ad\metadata.js:197:19)
    at Strategy.jwtVerify [as _verify] (xxx\server\node_modules\passport-azure-ad\lib\passport-azure-ad\bearerstrategy.js:176:36)
    at Strategy.authenticate (xxx\server\node_modules\passport-azure-ad\node_modules\passport-http-bearer\lib\strategy.js:132:10)
    at attempt (xxx\server\node_modules\passport\lib\middleware\authenticate.js:348:16)
    at authenticate (xxx\server\node_modules\passport\lib\middleware\authenticate.js:349:7)
    at Layer.handle [as handle_request] (xxx\server\node_modules\express\lib\router\layer.js:95:5)
    at trim_prefix (xxx\server\node_modules\express\lib\router\index.js:312:13)
    at xxx\server\node_modules\express\lib\router\index.js:280:7
    at Function.process_params (xxx\server\node_modules\express\lib\router\index.js:330:12)
    at next (xxx\server\node_modules\express\lib\router\index.js:271:10)
    at initialize (xxx\server\node_modules\passport\lib\middleware\initialize.js:53:5)
    at Layer.handle [as handle_request] (xxx\server\node_modules\express\lib\router\layer.js:95:5)
    at trim_prefix (xxx\server\node_modules\express\lib\router\index.js:312:13)
    at xxx\server\node_modules\express\lib\router\index.js:280:7
    at Function.process_params (xxx\server\node_modules\express\lib\router\index.js:330:12)
    at next (xxx\server\node_modules\express\lib\router\index.js:271:10)
```
Or some times it rejects valid json token, and messages, "An error was received validating the token secret or public key must be provided". I think it happens because at the time when token comes for validation the public key hasn't downloaded yet. For reproduce this issue I stop and start node server, and immediately after that make request.  On later requests with the same token there isn't any issue.
 Below is my node code.
```
if (env !== 'test') {
  app.use(passport.initialize());

  var authentication = require('./authentication/strategy');

[...]

PR #140, opened 2016-07-07
Solved #90: Cannot read property 'keys' of undefined

lovemaths:
Metadata loading is async, passport may use the metadata content even if the metadata loading is not finished, for instance, keys. In this fix, the following are done:

(1) load metadata dynamically at `authenticate` time, from server or memory cache
(2) solved async issue using waterfall
(3) added a couple of unit tests

### DIFF
--- a/lib/bearerstrategy.js
+++ b/lib/bearerstrategy.js
@@ -22,14 +22,21 @@
  */
 'use strict';
 
-// This strategy inherts from the passport-http-bearer strategy
-const BearerStrategyBase = require('passport-http-bearer').Strategy;
+/* eslint no-underscore-dangle: 0 */
 
-const util = require('util');
+const async = require('async');
+const cacheManager = require('cache-manager');
+const jws = require('jws');
 const jwt = require('jsonwebtoken');
+const passport = require('passport');
+const util = require('util');
+
 const Metadata = require('./metadata').Metadata;
 const Log = require('./logging').getLogger;
-const jws = require('jws');
+
+const log = new Log('AzureAD: Bearer Strategy');
+const memoryCache = cacheManager.caching({ store: 'memory', max: 3600, ttl: 1800 /* seconds */ });
+const ttl = 1800; // 30 minutes cache
 
 /**
  * Applications must supply a `verify` callback, for which the function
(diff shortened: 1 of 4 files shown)


lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
+ * be loaded dynamically (and cached) in the authentication time based on the request
+ * (2) async issue, metadata loading is async, passport-http-bearer's `authenticate`
+ * function depends on the metadata, and it might be called before the metadata loading
+ * finishes
+ *
+ * Therefore, what we do here is let the metadata loading happen in `authenticate`
+ * function, and use waterfall to make sure the authentication code runs after the
+ * metadata loading is finished. To do this, we write our own `authenticate` function
+ * and no longer inherit passport-http-bearer strategy.
+ */
```
polita: Is BearerStrategyBase still used anywhere?

lib/metadata.js
```diff
@@ -174,16 +174,21 @@ Metadata.prototype.updateOidcMetadata = function updateOidcMetadata(doc, next) {
...
+  if (!keys) {
+    throw new Error('keys is missing');
   }
 
   keys.some((key) => {
     log.info('working on key:', key);
 
     // are we working on the right key?
-    if (!key.kid === kid) {
+    if (key.kid !== kid) {
```
polita: Wasn't this fix in a different PR?

lib/bearerstrategy.js
lovemaths: No, it is no longer used.

lib/metadata.js
lovemaths: Yes, but I haven't got ship it for that PR yet. It would be great if that PR can be reviewed soon since it's important.

https://github.com/AzureAD/passport-azure-ad/pull/139

polita: Please don't check in the fix as part of this PR if you have it in another outstanding PR.

lib/bearerstrategy.js
polita: Please have @brandwe take a look as well.

brentschmaltz: Do we need to include what we 'used' to do in the comment?
Perhaps we could just say what we are doing.

lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
+Strategy.prototype.jwtVerify = function jwtVerifyFunc(req, token, done) {
+  const self = this;
 
-  if (opts.identityMetadata) {
-    log.info('Metadata url provided to Strategy was: ', opts.identityMetadata);
-    this.metadata = new Metadata(opts.identityMetadata, 'oidc', opts);
+  const decoded = jws.decode(token);
+  let PEMkey = null;
+
+  if (decoded == null) {
```
brentschmaltz: move this before 'let PEMkey = null'

lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
+    return done(null, false, 'Invalid JWT token.');
```
brentschmaltz: Is this an error condition? Should we throw as we do below?

lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
+  if (!options.validateIssuer) {
     log.warn(`We are not validating the issuer.
       This is fine if you are expecting multiple organizations to connect to your app.
       Otherwise you should validate the issuer.`);
   }
 
   // if you want to check JWT audience (aud), provide a value here
-  if (opts.audience) {
-    log.info('Audience provided to Strategy was: ', opts.audience);
+  if (options.audience) {
```
brentschmaltz: Checking audience is a mitigation against forwarding attacks. By default audience should always be validated.

lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
+      if (err instanceof jwt.JsonWebTokenError) {
+        log.warn('An error was received validating the token', err.message);
+        return done(null, false, util.format('Invalid token (%s)', err.message));
+      }
+      return done(err, false);
     }
-
-    if (opts.validateIssuer) {
-      opts.issuer = self.metadata.oidc.issuer;
+    log.info(verifiedToken, 'was token going out of verification');
```
brentschmaltz: I can't understand this log message.

lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
+   *      // result now equals 'done'    
+   * }); 
+   */
+  async.waterfall([
 
-    log.info('token decoded:  ', decoded);
+    // compute metadata url
+    (next) => {
+      if (!self._options.certificate && !self._options.identityMetadata) {
+        log.warn('No options was presented to Strategy as required.');
```
brentschmaltz: I think this log message should be something like: "options.certificate && options.identityMetadata are both null. It is not possible to validate a JWT token. These values supply 'keys' required for validating signatures."

lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
 
-    jwt.verify(token, PEMkey, options, (err, verifiedToken) => {
-      if (err) {
-        if (err instanceof jwt.TokenExpiredError) {
-          log.warn('Access token expired');
-          return done(null, false, 'The access token expired');
+      // default key for metadata cache
+      var cacheKey = 'ordinary';
+
+      var metadatURL = self._options.identityMetadata;
```
brentschmaltz: suggest changing variable name from _options.identityMetadata to: _options.metadataUrl.

brentschmaltz: why do we need the local variable: var metadatUrl (also spelling - add an 'a': metadataUrl)

lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
+
+      if (self._options.policyName) {
+        log.info('B2C: We have been instructed that this is a B2C tenant. We will configure as required.');
```
brentschmaltz: It would be helpful to log the 'policyName' that you are using.
Something like: log.info('B2C: using policy: %s', self._options.policyName);

lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
+        
+        if (!self._options.tenantName) {
+          return next(new TypeError('BearerStrategy requires you to pass the tenant name if using a B2C tenant.'));
```
brentschmaltz: Can we log all B2C logs as starting with 'B2C: details ...'

lib/bearerstrategy.js
```diff
@@ -118,121 +125,298 @@ const jws = require('jws');
...
         }
-        if (err instanceof jwt.JsonWebTokenError) {
-          log.warn('An error was received validating the token', err.message);
-          return done(null, false, util.format('Invalid token (%s)', err.message));
+
+        // We are replacing the common endpoint with the concrete metadata of a B2C tenant.
+        metadatURL = metadatURL
```
brentschmaltz: what happens if 'common' just happens to be in the user's url? suppose 'https://somehost/commonusersgohere' OR 'https://somehost/common/users/go/here'

lib/bearerstrategy.js
polita: I like this idea, but I recommend we use "CPIM" instead of "B2C".

lib/bearerstrategy.js
```diff
@@ -22,14 +22,21 @@
...
+const passport = require('passport');
 const util = require('util');
 const jwt = require('jsonwebtoken');
+const async = require('async');
+const jws = require('jws');
+
 const Metadata = require('./metadata').Metadata;
 const Log = require('./logging').getLogger;
-const jws = require('jws');
+const cacheManager = require('cache-manager');
```
brentschmaltz: move above.

lib/bearerstrategy.js
lovemaths: we make a copy of self._options.identityMetadata, later we will modify the copy (replacing 'common', and concatenate the policyName) and use the copy, we don't want to modify self._options.identityMetadata

lib/bearerstrategy.js
lovemaths: "B2C" is all over the places in this library, so let's stick to "B2C" for now.

lib/bearerstrategy.js
lovemaths: added, see below

lib/bearerstrategy.js
lovemaths: changed the one below to use "done(...)"

lib/bearerstrategy.js
lovemaths: done

lib/bearerstrategy.js
lovemaths: removed 'used to do' part

lib/bearerstrategy.js
lovemaths: changed

lib/bearerstrategy.js
lovemaths: done

lib/bearerstrategy.js
lovemaths: done

lib/bearerstrategy.js
```diff
@@ -117,122 +124,290 @@ const jws = require('jws');
...
 
   // if you want to check JWT audience (aud), provide a value here
-  if (opts.audience) {
-    log.info('Audience provided to Strategy was: ', opts.audience);
+  if (options.audience) {
+    log.info('Audience provided to Strategy was: ', options.audience);
+  } else {
+    log.warn(`We are not checking the audience because audience is not provided in the config file. \
+      Checking audience is a mitigation against forwarding attacks, providing the audience in the \
+      config file is strongly recommended for the sake of security.`);
```
brentschmaltz: just remove: 'for the sake of security' and :shipit:

PR #143, opened 2016-07-07
Fixed #135, wrong path in examples

### LINKED ISSUE #135: Running examples fails with Cannot find module '../../lib/passport-azure-ad/index'
Repro steps:

1. Clone the repo
2. Adjust the config.js files with proper value
3. Run the server.js or app.js

I have tried this with _login-oidc_ and _server-oauth2_, they all fail with the same error. Seems like the ```require``` path is wrong?

lovemaths:
should be '../../lib/index', not '../../lib/passport-azure-ad/index'.


polita:
:shipit:

Issue #91, opened 2016-03-23
TypeError: this.fail is not a function


julianfrank:
I'm having a NodeJS App and am using your module to integrate authentication with azure AD...Everything is cozy with 'assigned' users. 

I get this error whenever I try to login into my app with an 'Un-Assigned User' ...As per Passport documentation this should route the page back to login page but in my app this error is thrown and the application exits!...Could you please suggest how I should work around this?

**Here is the dump in NodeJS:**
TypeError: this.fail is not a function
    at d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\passport-azure-ad\lib\passport-azure-ad\oidcstrategy.js:174:17
    at fn (d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\async\lib\async.js:746:34)
    at d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\async\lib\async.js:1213:16
    at d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\async\lib\async.js:166:37
    at d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\async\lib\async.js:706:43
    at d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\async\lib\async.js:167:37
    at d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\async\lib\async.js:1209:30
    at d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\passport-azure-ad\lib\passport-azure-ad\oidcstrategy.js:157:10
    at d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\passport-azure-ad\lib\passport-azure-ad\oidcstrategy.js:735:13
    at d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\async\lib\async.js:52:16
    at Immediate.<anonymous> (d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\async\lib\async.js:1206:34)
    at Immediate.wrapped [as _onImmediate] (d:\Users\Johan\Documents\CloudCode\jfwrtc\node_modules\newrelic\lib\transaction\tracer\index.js:155:28)
    at processImmediate [as _immediateCallback] (timers.js:516:17)

**Here is the POST I receive from Azure AD**...Which is as per expectation for un-assigned users:
[...]

PR #149, opened 2016-07-19
Issue #91, TypeError: this.fail is not a function

lovemaths:
When there is 'error' in the response, passport should call this.fail instead of this.error. The difference is:
(1) this.error is for server error
(2) this.fail is for authentication error, where the server has no problem. In this case, the server can redirect the user to login page (or whatever page the developer sets) for authentication.

### DIFF
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -210,21 +210,27 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
       },
       // handle error response
       (next) => {
-        if (!(req.query && req.query.code)) {
+        var err = undefined;
+        var err_description = undefined;
+
+        if (req.query && req.query.error) {
+          err = req.query.error;
+          err_description = req.query.error_description;
+        } else if (req.body && req.body.error) {
+          err = req.body.error;
+          err_description = req.body.error_description;
+        } else {
           return next();
         }
-        // @TODO: need more specific error handling
-        //        response might include error_code, error_details and more
 
-        // Error information pertaining to OAuth 2.0 flows is encoded in the
-        // query parameters, and should be propagated to the application.
+        log.info('Error received in the response was: ', err);
+        if (err_description)
+          log.info('Error description received in the response was: ', err_description);
 
-        log.info(`OPENID CONNECT ERROR: We are in the OpenID Connect Response with error.
-          Assumed because no auth code was in the query.`);
-        log.info('Error received was: ', req.query.error);
-
-        // pass error to `next` and handle error occurance centrally in `async.waterfall`'s callback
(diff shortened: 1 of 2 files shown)


lib/oidcstrategy.js
```diff
@@ -210,21 +210,27 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
 
-        log.info(`OPENID CONNECT ERROR: We are in the OpenID Connect Response with error.
-          Assumed because no auth code was in the query.`);
-        log.info('Error received was: ', req.query.error);
-
-        // pass error to `next` and handle error occurance centrally in `async.waterfall`'s callback
-        return next(req.query.error);
+        // Unfortunately, we cannot return the 'error description' to the user, since 
+        // it goes to http header by default and it usually contains characters that
+        // http header doesn't like, which causes the program to crash.
```
polita: Should we be URL encoding the description so it can be in the header without crashing? Is it possible clients are expecting the error description in the header to be shown to the user?

lib/oidcstrategy.js
```diff
@@ -210,21 +210,27 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
+          err_description = req.body.error_description;
+        } else {
           return next();
         }
-        // @TODO: need more specific error handling
-        //        response might include error_code, error_details and more
 
-        // Error information pertaining to OAuth 2.0 flows is encoded in the
-        // query parameters, and should be propagated to the application.
+        log.info('Error received was: ', err);
```
polita: If this is a server error sent to us, can you add something to the log that makes that clear? Maybe "Server error received was: "...

lovemaths: This is not a server error. I can change it to "log.info('Error received in the response was: ..." if that's more clear.

polita: Sounds good.

lib/oidcstrategy.js
lovemaths: Normally developer won't expect URL encoded error message, they just want to a normal string. There is already an issue filed on this (https://github.com/AzureAD/passport-azure-ad/issues/104), we don't have to solve it here.

lib/oidcstrategy.js

lib/oidcstrategy.js
polita: Does this checkin remove error info from the header?

lovemaths: The original code just throws an error and doesn't put anything in the header. This fix calls the correct function (this.fail) and adds 'error' info into the header, calling this.fail enables developer to redirect the user to the login page for authentication.

polita:
:shipit:

PR #150, opened 2016-07-21
work for Jenkins policheck

lovemaths:
Jenkins policheck script doesn't like the word 'foo', we are replacing 'foo' with the word 'something' so the policheck can pass.


polita:
:shipit:

tushargupta51:
:shipit:

Issue #146, opened 2016-07-13
potential duplicate use of "new Buffer(*, 'base64')"


lovemaths:
Need to investigate

in metadata.js,   function Metadata.prototype.generateOidcPEM
    const modulus = new Buffer(key.n, 'base64');   <=======
    const exponent = new Buffer(key.e, 'base64');  <=======

    pubKey = aadutils.rsaPublicKeyPem(modulus, exponent);  <=== pass to rasPublicKeypem function

in aadutils.js

exports.rsaPublicKeyPem = (modulusB64, exponentB64) => {
  const modulus = new Buffer(modulusB64, 'base64');      <== modulusB64 is already a 'base64' buffer
  const exponent = new Buffer(exponentB64, 'base64');   <== exponentB64 is already a 'base64' buffer

lovemaths:
done

PR #151, opened 2016-07-21
Issue #146, duplicate use of "new Buffer"

lovemaths:
When we have key.n or key.e, we turn it into base64 encoded buffer
```
    var modulus64 = new Buffer(key.n, 'base64');
```
Later we feed it to rsaPublicKeyPem function, which does:
```
    var modulus =  new Buffer(modulus64, 'base64');
```

Here modulus and modulus64 are the same. Since Buffer(buffer) only takes 1 parameter, 'base64' will be ignored, and this function just makes a copy of modulus. Therefore, we just need to feed key.n or key.e to rasPublicKeyPem instead of turning it into buffer first.

### DIFF
--- a/lib/metadata.js
+++ b/lib/metadata.js
@@ -205,10 +205,7 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
     }
 
     // generate PEM from `modulus` and `exponent`
-    const modulus = new Buffer(key.n, 'base64');
-    const exponent = new Buffer(key.e, 'base64');
-
-    pubKey = aadutils.rsaPublicKeyPem(modulus, exponent);
+    pubKey = aadutils.rsaPublicKeyPem(key.n, key.e);
     foundKey = true;
     
     return pubKey;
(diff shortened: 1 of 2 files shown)


polita:
> Since Buffer(buffer) only takes 1 parameter, 'base64' will be ignored, and this function just makes a copy of modulus.

I don't think that's right. There's a Buffer ctor overload that takes the encoding (although all these ctors are deprecated). See https://nodejs.org/api/buffer.html#buffer_new_buffer_str_encoding. Can you change all our calls to the ctor to the recommended Buffer.from or Buffer.alloc?

lovemaths:
@polita I think that's right. The one with encoding is:  Buffer(string, encoding), and there is no Buffer(buffer, encoding) ctor. I have tested it as well. The following are the ctors from that website:
  new Buffer(array)
  new Buffer(buffer)
  new Buffer(arrayBuffer[, byteOffset [, length]])
  new Buffer(size)
  new Buffer(str[, encoding])

Yes. I will change them to Buffer.from.

polita:
Got it. I couldn't tell the type of key.n. Thanks for movin to the new APIs.

lovemaths:
@polita Buffer.from(str, encoding) is introduced in v5.10.0, Buffer ctors are deprecated from v6.0.0, however, the recommended version for most developers to use is still v4.4.7 (see https://nodejs.org/en/), so we still need to use these ctors.

polita:
Ok, seems reasonable. :shipit:

PR #152, opened 2016-07-22
Typo in Gruntfile.js

lovemaths:
Folder name should be `Chai-passport_test` instead of `chai-passport_test`. Tests in this folder run locally, but not on travis, seems my version of npm doesn't distinguish the lower/upper case.


polita:
:shipit:

Issue #71, opened 2015-10-19
Cryptic error message when the client ID is null/undefined


jeffwilcox:
Hi,
Might be useful to actively protect against a null or undefined/empty client ID when using the module.

This is the cryptic error I get when I set client ID to the empty string, for example...

```
...\node_modules\passport-azure-ad\node_modules\cache-manager\lib\callback_filler.js:13
    waiting.forEach(function(task) {
           ^

TypeError: Cannot read property 'forEach' of undefined
    at CallbackFiller.fill (...\node_modules\passport-azure-ad\node_modules\cache-manager\lib\callback_filler.js:13:12)
    at Domain.<anonymous> (...\node_modules\passport-azure-ad\node_modules\cache-manager\lib\caching.js:85:36)
    at emitOne (events.js:77:13)
    at Domain.emit (events.js:169:7)
    at emitError (domain.js:65:24)
    at Domain.errorHandler [as _errorHandler] (domain.js:104:16)
    at process._fatalException (node.js:218:33)
    at Domain.errorHandler [as _errorHandler] (domain.js:130:26)
    at process._fatalException (node.js:218:33)
```

brandwe:
This is damn strange. I wrote a validator class to do this ahead of time. I assume it's being used before that time. I'll fix.

lovemaths:
done

PR #154, opened 2016-07-26
Issue #71

lovemaths:
Some properties (clientID, callbackURL, responseType, responseMode) from the provided options are checked in the `authenticate` time, this is not good because:
(1) code should crash immediately instead of the time when `authenticate` is called, immediate crash gives the user immediate feedback of their configuration error
(2) these properties are checked for every request that needs `authenticate`. This is not necessary since the value of these properties would be changed.
Considering all these, we now check them at the creation time of strategy.

### DIFF
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -143,6 +143,22 @@ function Strategy(options, verify) {
       This is fine if you are expecting multiple organizations to connect to your app.
       Otherwise you should validate the issuer.`);
   }
+
+  // validate other necessary option items provided, we validate them here and only once
+  var itemsToValidate = objectTransform({
+    source: options,
+    pick: ['clientID', 'callbackURL', 'responseType', 'responseMode']
+  });
+  var validatorConfiguration = {
+    clientID: Validator.isNonEmpty,
+    callbackURL: Validator.isNonEmpty,
+    // callbackURL: Validator.isURL  // We allow http for now
+    responseType: Validator.isTypeLegal,
+    responseMode: Validator.isModeLegal
+  };
+  // validator will throw exception if a required option is missing
+  var validator = new Validator(validatorConfiguration);
+  validator.validate(itemsToValidate);
 }
 
 // Inherit from `passport.Strategy`.
@@ -619,13 +635,8 @@ Strategy.prototype.setOptions = function setOptions(options, metadataUrl, done) 
 
       // Now that we have our options for configuration, let's check them for issues.
       const validatorConfig = {
-        clientID: Validator.isNonEmpty,
-        callbackURL: Validator.isNonEmpty,
-        responseType: Validator.isTypeLegal,
-        responseMode: Validator.isModeLegal,
         authorizationURL: Validator.isURL,
         tokenURL: Validator.isURL,
-        // callbackURL: Validator.isURL  // We allow http for now :-/
       };
 
       // validator will throw exception if a required option is missing
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -147,14 +147,14 @@ function Strategy(options, verify) {
   // validate other necessary option items provided, we validate them here and only once
   var itemsToValidate = objectTransform({
     source: options,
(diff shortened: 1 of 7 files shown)


lib/oidcstrategy.js
```diff
@@ -143,6 +143,22 @@ function Strategy(options, verify) {
...
+  var validatorConfiguration = {
+    clientID: Validator.isNonEmpty,
+    callbackURL: Validator.isNonEmpty,
+    // callbackURL: Validator.isURL  // We allow http for now
+    responseType: Validator.isTypeLegal,
+    responseMode: Validator.isModeLegal
+  };
+  // validator will throw exception if a required option is missing
+  var validator = new Validator(validatorConfiguration);
+  validator.validate(itemsToValidate);
```
polita: Can the config be changed after initialization? If so, the configuration needs to be checked as close to usage of the configuration as possible.

lib/oidcstrategy.js
```diff
@@ -143,6 +143,22 @@ function Strategy(options, verify) {
...
+
+  // validate other necessary option items provided, we validate them here and only once
+  var itemsToValidate = objectTransform({
+    source: options,
+    pick: ['clientID', 'callbackURL', 'responseType', 'responseMode']
+  });
+  var validatorConfiguration = {
+    clientID: Validator.isNonEmpty,
+    callbackURL: Validator.isNonEmpty,
+    // callbackURL: Validator.isURL  // We allow http for now
```
polita: Why is this validation commented out?

lib/oidcstrategy.js
```diff
@@ -143,6 +143,22 @@ function Strategy(options, verify) {
...
+    responseType: Validator.isTypeLegal,
+    responseMode: Validator.isModeLegal
```
polita: Should we be validating the identityMetadata?

lovemaths: It's at line 130.

lib/oidcstrategy.js
lovemaths: isURL only allows https, not http

lib/oidcstrategy.js
lovemaths: No, config won't be changed.

polita: What happens if the user tries to change config after init? Do they get an error?

lib/oidcstrategy.js
polita: Well, then isURL is poorly named. :) Is there a validation property that will validate that it's a URL that can be http or https?

lib/oidcstrategy.js
polita: Line 130 validates that it's set. Do we validate that it's a well-formed URL?

lib/oidcstrategy.js
jeffwilcox: Please do allow us to keep using HTTP for local dev scenarios without having to do a TLS dance; but a configuration property is ideal - for live dev, test and production scenarios we'd want to make sure that is off.

Real case is: give me `localhost:3000` please :).

lib/oidcstrategy.js
lovemaths: There is no way a user can change the config (clientID, callbackURL, responseType, responseMode values)  without restarting the program. The program needs recompilation to take the new values into effect.

lib/oidcstrategy.js
lovemaths: Don't have such a thing yet. I will write one.

lib/oidcstrategy.js
polita: Ok, great.

lib/oidcstrategy.js
lovemaths: Done. Now we have isHttpURL, isHttpsURL and isURL. For callbackURL, http is allowed so developer can use localhost for local dev; other URLs belongs to AAD so for those we only allow https.

lib/oidcstrategy.js
lovemaths: Done.

polita:
:shipit:

Issue #157, opened 2016-07-28
Robustness work for OIDCStrategy


brandwe:
The id_token should be verified before the profile is loaded and handed back to the developer.

dennisx5:
I suppose the failureRedirect should be used as to route to, when id_token is not present or validated to false. Currently at my returnUrl I only receive the errormessage "Error: ID Token not present in response" and can't navigate back to my app.

jeffwilcox:
@dennisx5 thank you - I've been seeing that specific error message for a while and super confused when that has showed up in logs or error reports.

PR #161, opened 2016-08-09
Issue #111 and #157  Add security checks for oidcstrategy

lovemaths:
Breaking change: response_type = 'query' is not longer allowed, since according to the openid spec, query must not be used for hybrid and implicit flow. Now we uniformly use response_type = 'form_post'.

Added:
(1) id_token and nonce validation
(2) code, c_hash validation
(3) access_token, at_hash validation
(4) state validation

### DIFF
--- a/lib/aadutils.js
+++ b/lib/aadutils.js
@@ -22,6 +22,7 @@
  */
 'use strict';
 
+const base64url = require('base64url');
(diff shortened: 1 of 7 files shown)


lovemaths:
will remove the breaking change and do a new PR

PR #162, opened 2016-08-10
Issue #111 and #157  Oidcstrategy robustness

lovemaths:
Added:
(1) id_token and nonce validation
(2) code, c_hash validation
(3) access_token, at_hash validation
(4) state validation

### DIFF
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -143,6 +143,22 @@ function Strategy(options, verify) {
       This is fine if you are expecting multiple organizations to connect to your app.
       Otherwise you should validate the issuer.`);
   }
+
+  // validate other necessary option items provided, we validate them here and only once
+  var itemsToValidate = objectTransform({
+    source: options,
+    pick: ['clientID', 'callbackURL', 'responseType', 'responseMode', 'identityMetadata']
+  });
+  var validatorConfiguration = {
+    clientID: Validator.isNonEmpty,
+    callbackURL: Validator.isURL,   // allow http so developer can use http://localhost:3000
+    responseType: Validator.isTypeLegal,
+    responseMode: Validator.isModeLegal,
+    identityMetadata: Validator.isHttpsURL
+  };
+  // validator will throw exception if a required option is missing
+  var validator = new Validator(validatorConfiguration);
+  validator.validate(itemsToValidate);
 }
 
 // Inherit from `passport.Strategy`.
@@ -619,13 +635,8 @@ Strategy.prototype.setOptions = function setOptions(options, metadataUrl, done) 
 
       // Now that we have our options for configuration, let's check them for issues.
       const validatorConfig = {
-        clientID: Validator.isNonEmpty,
-        callbackURL: Validator.isNonEmpty,
-        responseType: Validator.isTypeLegal,
-        responseMode: Validator.isModeLegal,
-        authorizationURL: Validator.isURL,
-        tokenURL: Validator.isURL,
-        // callbackURL: Validator.isURL  // We allow http for now :-/
+        authorizationURL: Validator.isHttpsURL,
+        tokenURL: Validator.isHttpsURL,
       };
 
       // validator will throw exception if a required option is missing
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -23,6 +23,8 @@
 
 'use strict';
 
+const UrlValidator = require('valid-url');
+
 const types = {};
 
 function Validator(config) {
@@ -77,16 +79,24 @@ types.isModeLegal = {
 
 Validator.isURL = 'isURL';
 types.isURL = {
-
   validate: (value) => {
-    const pattern = new RegExp('^(https?:\\/\\/)?' + // protocol
-      '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' + // domain name
-      '((\\d{1,3}\\.){3}\\d{1,3}))' + // OR ip (v4) address
-      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port and path
-      '(\\?[;&a-z\\d%_.~+=-]*)?' + // query string
-      '(\\#[-a-z\\d_]*)?$', 'i'); // fragment locator
+    return UrlValidator.isHttpUri(value) || UrlValidator.isHttpsUri(value);
+  },
+  error: 'The URL must be valid and be https:// or http://',
+};
 
-    return !!pattern.test(value);
+Validator.isHttpURL = 'isHttpURL';
+types.isHttpURL = {
+  validate: (value) => {
+    return UrlValidator.isHttpUri(value);
+  },
+  error: 'The URL must be valid and be http://',
+};
+
+Validator.isHttpsURL = 'isHttpsURL';
+types.isHttpsURL = {
+  validate: (value) => {
+    return UrlValidator.isHttpsUri(value);
   },
   error: 'The URL must be valid and be https://',
 };
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "request": "^2.72.0",
     "underscore": "^1.8.3",
     "webfinger": "^0.4.2",
+    "valid-url": "^1.0.6",
     "xml-crypto": "^0.8.4",
     "xml2js": "^0.4.16",
     "xmldom": "^0.1.22",
--- a/test/Chai-passport_test/oidc_error_test.js
+++ b/test/Chai-passport_test/oidc_error_test.js
@@ -33,13 +33,13 @@ chai.use(require('chai-passport-strategy'));
 
 // Mock options required to create a OIDC strategy
 var options = {
-    callbackURL: '/returnURL',
+    callbackURL: 'http://returnURL',
     clientID: 'my_client_id',
     clientSecret: 'my_client_secret',
-    identityMetadata: 'www.example.com/metadataURL',
+    identityMetadata: 'https://www.example.com/metadataURL',
     skipUserProfile: true,
-    responseType: 'form_post',
-    responseMode: 'id_token',
+    responseType: 'id_token',
+    responseMode: 'form_post',
     validateIssuer: true,
     passReqToCallback: false,
     sessionKey: 'my_key'    //optional sessionKey
@@ -54,8 +54,8 @@ testStrategy.configure = function(identifier, done) {
   var opt = {           
     clientID: options.clientID,
     clientSecret: options.clientSecret,
-    authorizationURL: 'www.example.com/authorizationURL',
-    tokenURL: 'www.example.com/tokenURL'
+    authorizationURL: 'https://www.example.com/authorizationURL',
+    tokenURL: 'https://www.example.com/tokenURL'
   };
   done(null, opt);
 };
--- a/test/Nodeunit_test/oidc_b2c_test.js
+++ b/test/Nodeunit_test/oidc_b2c_test.js
@@ -60,6 +60,10 @@ exports.oidc = {
       identityMetadata: 'https://login.microsoftonline.com/common/.well-known/openid-configuration',
       tenantName: 'hypercubeb2c.onmicrosoft.com',
       forceB2C: true,
+      clientID: '123',
+      callbackURL: 'http://www.example.com',
+      responseType: 'id_token',
+      responseMode: 'form_post'
     };
 
     test.doesNotThrow(
--- a/test/Nodeunit_test/oidc_test.js
+++ b/test/Nodeunit_test/oidc_test.js
@@ -85,54 +85,105 @@ exports.oidc = {
 
     test.done();
   },
-  'with missing option resposneType': (test) => {
+  'with invalid option clientID': (test) => {
     test.expect(1);
     // tests here
 
     const oidcConfig = {
       // required options
       identityMetadata: 'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration',
-      responseType: 'id_tokennn', // for login only flows use id_token. For accessing resources use `id_token code`
+      clientID: '',  // invalid
+      callbackURL: 'http://www.example.com',
+      responseType: 'id_token',
+      responseMode: 'form_post'
     };
     test.throws(
       () => {
         const s = new OidcStrategy(oidcConfig, noop);
-        s.loadOptions(oidcConfig, noop);
       },
       TypeError,
-      'Should fail with wrong reponses config options'
+      'Should fail with wrong response config options'
     );
 
     test.done();
   },
-  'with missing option resposneMode': (test) => {
+  'with invalid option callbackURL': (test) => {
     test.expect(1);
     // tests here
 
     const oidcConfig = {
       // required options
       identityMetadata: 'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration',
-      responseMode: 'fragment', // For login only flows we should have token passed back to us in a POST
+      clientID: '123',
+      callbackURL: '',  // invalid
+      responseType: 'id_tokennn',
+      responseMode: 'form_post'
     };
     test.throws(
       () => {
         const s = new OidcStrategy(oidcConfig, noop);
-        s.loadOptions(oidcConfig, noop);
       },
-      Error,
-      'Should fail with wrong reponses config options'
+      TypeError,
+      'Should fail with wrong response config options'
     );
 
     test.done();
   },
-  'with options': (test) => {
+  'with invalid option responseType': (test) => {
     test.expect(1);
     // tests here
 
     const oidcConfig = {
       // required options
-      identityMetadata: 'https://login.microsoftonline.com/common/.well-known/openid-configuration',
-      issuer: 'http://localhost:3000', // this is the URI you entered for APP ID URI when configuring SSO for you app on Azure AAD
+      identityMetadata: 'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration',
+      clientID: '123',
+      callbackURL: 'http://www.example.com',
+      responseType: 'id_tokennn', // invalid
+      responseMode: 'form_post'
+    };
+    test.throws(
+      () => {
+        const s = new OidcStrategy(oidcConfig, noop);
+      },
+      TypeError,
+      'Should fail with wrong response config options'
+    );
+
+    test.done();
+  },
+  'with invalid option responseMode': (test) => {
+    test.expect(1);
+    // tests here
+
+    const oidcConfig = {
+      // required options
+      identityMetadata: 'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration',
+      clientID: '123',
+      callbackURL: 'http://www.example.com',
+      responseType: 'id_token',
+      responseMode: 'fragment' // invalid
+    };
+    test.throws(
+      () => {
+        const s = new OidcStrategy(oidcConfig, noop);
+      },
+      Error,
+      'Should fail with wrong response config options'
+    );
+
+    test.done();
+  },
+  'with valid options': (test) => {
+    test.expect(1);
+    // tests here
+
+    const oidcConfig = {
+      // required options
+      identityMetadata: 'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration',
+      clientID: '123',
+      callbackURL: 'http://www.example.com',
+      responseType: 'id_token',
+      responseMode: 'form_post'
     };
 
     test.doesNotThrow(
--- a/test/Nodeunit_test/oidc_v2_test.js
+++ b/test/Nodeunit_test/oidc_v2_test.js
@@ -73,6 +73,8 @@ exports.oidc = {
     const oidcConfig = {
       // required options
       identityMetadata: 'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration',
+      clientID: '123',
+      callbackURL: 'http://www.example.com',
       responseType: 'id_token', // for login only flows use id_token. For accessing resources use `id_token code`
       responseMode: 'form_post', // For login only flows we should have token passed back to us in a POST
     };


lib/nonceHandler.js
```diff
@@ -0,0 +1,55 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+ * OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT
+ * OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+ 'use restrict';
+
+ exports.verifyNonce = function(req, sessionKey, nonce_in_idtoken) {
+	if (!req.session) {
```
addev-ashish: >	 [](start = 0, length = 1)

file has tabs, please change to spaces

lib/stateHandler.js
```diff
@@ -0,0 +1,66 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+ * OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT
+ * OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+ 
+'use strict';
+
+exports.verifyState = function(req, sessionKey) {
+	if (!req.session) {
```
addev-ashish: >	if [](start = 0, length = 3)

i understand exisiting code, but please avoid tabs in any new code we write.

lib/nonceHandler.js
```diff
@@ -0,0 +1,55 @@
...
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+ * OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT
+ * OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+ 'use restrict';
+
+ exports.verifyNonce = function(req, sessionKey, nonce_in_idtoken) {
```
addev-ashish: >nonce_in_idtoken [](start = 49, length = 16)

you do not need nonce_in_idtoken as a param, it its false you should not call this function. The only way this function succeeds is if nonce_in_idtoken is true and the rest of your checks pass.

lib/oidcstrategy.js
```diff
@@ -172,6 +219,44 @@ util.inherits(Strategy, passport.Strategy);
  * @api protected
  */
 Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
+  /* 
+   * We should be careful using 'this'. Avoid the usage like `this.xxx = ...`
+   * unless you know what you are doing.
```
RandalliLama: Is this a generic comment on JavaScript, or something specific to the OIDC code.  If it is the former I would remove it.

lib/aadutils.js
```diff
@@ -155,3 +156,21 @@ exports.rsaPublicKeyPem = (modulusB64, exponentB64) => {
 
   return pem;
 };
+
+// used for c_hash and at_hash validation
+// case (1): content = access_token, hashProvided = at_hash
+// case (2): content = code, hashProvided = c_hash
+exports.checkHashValueRS256 = (content, hashProvided) => {
+  // step 1. hash the content
+  var digest = crypto.createHash('sha256').update(content).digest();
```
polita: Is there any case where this can return null?

lib/oidcstrategy.js
lovemaths: This is specific to OIDC code.

lib/nonceHandler.js
polita: @addev-ashish, I think nonce_in_idtoken is the actual nonce, not a boolean. @lovemaths, is that right?

lovemaths: @polita @addev-ashish yes, the nonce_in_idtoken is the actual nonce, not a boolean

lib/oidcstrategy.js
```diff
@@ -579,21 +401,21 @@ Strategy.prototype.setOptions = function setOptions(options, metadataUrl, done)
     function loadMetadata(next) {
       log.info('Parsing Metadata: ', metadataUrl);
 
-      memoryCache.wrap(self._cacheKey, (cacheCallback) => {
-        const metadata = new Metadata(metadataUrl, 'oidc', options);
-        metadata.fetch((fetchMetadataError) => {
+      memoryCache.wrap(cachekey, (cacheCallback) => {
+        //self.metadata = new Metadata(metadataUrl, 'oidc', options);
```
polita: Why is this commented out?

lib/oidcstrategy.js
```diff
@@ -752,4 +574,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+  log.info('token decoded: ', decoded);
+
+  // get Pem Key
+  var PEMkey = undefined;
+  if (decoded.header.x5t) {
+    PEMkey = self.metadata.generateOidcPEM(decoded.header.x5t);
+  } else if (decoded.header.kid) {
+    PEMkey = self.metadata.generateOidcPEM(decoded.header.kid);
+  } else {
+    return self.fail('We did not reveive a token we know how to validate');
```
polita: "receive"

RandalliLama:
There is a file 'TokenValidator.js'.  Is this file necessary at all.  It contains a version of generateOidcPEM that is different from the one in metadata.js.  If we can remove TokenValidator.js let's do it.

lib/oidcstrategy.js
```diff
@@ -752,4 +574,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+    // check c_hash
+    if (code && jwtClaims.c_hash && !utils.checkHashValueRS256(code, jwtClaims.c_hash))
+      return self.fail("invalid c_hash");
+
+    // check at_hash
+    if (access_token && jwtClaims.at_hash && !utils.checkHashValueRS256(access_token, jwtClaims.at_hash))
+      return self.fail("invalid at_hash");
+
+    // return jwt claims and jwt claims string
+    var idTokenSegments = id_token.split('.');
```
polita: We should verify at least that split returned 3 strings in the array.

lib/oidcstrategy.js
```diff
@@ -752,4 +574,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+              const userInfoURL = url.format(parsedUrl);
+
+              // ask oauth2 to use authorization header to bearer access token
+              oauth2.useAuthorizationHeaderforGET(true);
+              return oauth2.get(userInfoURL, access_token, (getUserInfoError, body) => {
+                if (getUserInfoError) {
+                  return next(new InternalOAuthError('failed to fetch user profile', getUserInfoError));
+                }
+
+                log.info('PROFILE LOADED FROM MS IDENTITY', body);
```
polita: Why is this all caps?

lib/nonceHandler.js
addev-ashish: but then how is nonce_provided = req.session[sessionKey] && req.session[sessionKey].nonce;

supposed to match it?

---
In reply to: [74174510](https://github.com/AzureAD/passport-azure-ad/pull/162#discussion_r74174510) [](ancestors = 74174510)

lib/oidcstrategy.js
```diff
@@ -752,4 +574,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+
+                var userinfoReceived = undefined;
+                // use try / catch around JSON.parse --> could fail unexpectedly
+                try {
+                  userinfoReceived = JSON.parse(body);
+                } catch (ex) {
+                  return next(ex);
```
polita: Are we logging this error?

lib/oidcstrategy.js
```diff
@@ -752,4 +574,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+    if (configureError) {
+      return next(configureError);
+    }
+
+    var options = self._options;
+
+    let callbackURL = options.callbackURL || config.callbackURL;
+    if (callbackURL) {
+      const parsed = url.parse(callbackURL);
+      if (!parsed.protocol) {
```
addev-ashish: >protocol [](start = 18, length = 8)

should you check if the protocol is https

lib/nonceHandler.js
lovemaths: @addev-ashish  && is sometimes called a guard operator.

variable = indicator && value
it can be used to set the value only if the indicator is truthy.

addev-ashish: ok

---
In reply to: [74176020](https://github.com/AzureAD/passport-azure-ad/pull/162#discussion_r74176020) [](ancestors = 74176020)

lib/oidcstrategy.js
lovemaths: @addev-ashish callbackURL is checked at the very beginning when strategy is created

lib/aadutils.js
lovemaths: I don't think so.

lib/oidcstrategy.js
lovemaths: Will remove this line.

lib/oidcstrategy.js
lovemaths: ok

lib/oidcstrategy.js
lovemaths: jwt.verify already checked it

lib/oidcstrategy.js
lovemaths: Will change it to 'Profile loaded from MS identity'

lib/oidcstrategy.js
lovemaths: program will crash and show the exception content, there is no need to log it.

lib/oidcstrategy.js
```diff
@@ -187,10 +272,16 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
+       * Step 1. compute metadata url 
+       */
       (next) => {
         // B2C interception
-        let metadata = self._options.identityMetadata;
+        let metadataUrl = self._options.identityMetadata;
+
+        // use 'ordinary' as the default cachekey, in the case of B2C, we will
+        // change the value to policy later
+        let cachekey = 'ordinary';
```
RandalliLama: Can we make the values 'ordinary' and 'policy' constants defined earlier in the file so that it is clearer what all of the options are for the cachekey?

lovemaths: @RandalliLama Yes 'TokenValidator.js' is no longer used, will remove it from the lib

RandalliLama:
@lovemaths The option being checked at the following line doesn't seem to do anything at all.  If I am right, then we should file an issue on this.

https://github.com/AzureAD/passport-azure-ad/blob/04ea99fce9387b60ea1d8a585f54cd37132ec1bd/lib/oidcstrategy.js#L201

lovemaths:
@RandalliLama The thing it does is to give the developer a warning that they should let the library validate the issuer. Developer has a choice not to validate the issuer for the multi-tenant scenario.

lib/oidcstrategy.js
lovemaths: @RandalliLama This way will be more stable, in the future if there is any change in the 'policy' values, I don't have to go back to change or add new constants.

lib/aadutils.js
```diff
@@ -22,6 +22,7 @@
  */
 'use strict';
 
+const base64url = require('base64url');
 const crypto = require('crypto');
```
tushargupta51: Is this the crypto library: https://github.com/Gozala/crypto? That repo does not look like supported at all. I know you did not add this but am not sure if we want to refer to this library.

npm link :https://www.npmjs.com/package/crypto

lovemaths: @polita @RandalliLama @aashiman @tushargupta51 No, crypto is the official one from nodejs (https://nodejs.org/api/crypto.html#crypto_crypto)

tushargupta51: Never mind. It is the node js inbuilt library: https://nodejs.org/api/crypto.html.

lib/oidcstrategy.js
```diff
@@ -752,4 +574,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+    return self.fail('We did not reveive a token we know how to validate');
+  }
+
+  var options = self._options;
+
+  // since the id_token is for itself, so the id_token is the clientID by default
+  options.audience = options.clientID;
+
+  // verify id_token signature and claims
+  return jwt.verify(id_token, PEMkey, options, (err, jwtClaims) => {
```
RandalliLama: @lovemaths I don't see anything that guarantees that an issuer was set in the options.  Maybe I missed it.  I would prefer that we enforce that here, unless issuer validation has been explicitly turned off, so there is no mistake.

RandalliLama:
@lovemaths I understand the intent of the options.validateIssuer field, but I don't see where it is being used.  Are the options being passed to some other library that is making use of that option.  If so, that should be made more obvious.  We should also ensure that the default is true.

lib/nonceHandler.js
tushargupta51: @lovemaths We should fail when nonce_in_idToken is null. As we always include the nonce in the outgoing request, one must be present in the response.

http://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDTValidation

lib/oidcstrategy.js
```diff
@@ -752,4 +573,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+ * validate id_token, and pass the validated claims and the payload to callback
+ * if code (resp. access_token) is provided, we will validate the c_hash (resp at_hash) as well
+ *
+ * @param {String} id_token
+ * @param {Object} req
+ * @param {Function} callback
+ * @param {String} code (if you want to validate c_hash)
+ * @param {String} access_token (if you want o validate at_hash)
+ */
+Strategy.prototype._validateIdToken = function validateIdToken(id_token, code, access_token, req, callback) {
```
tushargupta51: This method does more than id token validation. It validates code and access_token also in case of hybrid flows. Maybe validateResponse better name?

lib/oidcstrategy.js
```diff
@@ -752,4 +573,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+
+    return oauth2.getOAuthAccessToken(code, {
+      grant_type: 'authorization_code',
+      redirect_uri: callbackURL,
+      }, (getOAuthAccessTokenError, access_token, refresh_token, params) => {
+        if (getOAuthAccessTokenError) {
+          return next(new InternalOAuthError('failed to obtain access token', getOAuthAccessTokenError));
+        }
+
+        var id_token = params.id_token;
```
tushargupta51: When using 'code id_token' flow, redeeming code at token endpoint returns a second id_token. We must ensure that the `sub` and `iss` claims in this second token are same as the first token. http://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken2

lib/oidcstrategy.js
```diff
@@ -752,4 +573,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+    if (Array.isArray(jwtClaims.aud) && jwtClaims.aud.length > 1)
+      return self.fail('we do not allow multiple audiences in id_token');
+
+    // check the nonce in claims
+    var nonceCheckResult = nonceHandler.verifyNonce(req, self._key, jwtClaims.nonce);
+    if (!nonceCheckResult.valid)
+      return self.fail(nonceCheckResult.errorMessage);
+
+    // check c_hash
+    if (code && jwtClaims.c_hash && !utils.checkHashValueRS256(code, jwtClaims.c_hash))
```
tushargupta51: c_hash is not optional when using `code id_token` flow. If jwtClaims.c_hash is null, we should fail. http://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken

lib/oidcstrategy.js
```diff
@@ -752,4 +573,468 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+      return self.fail("invalid c_hash");
+
+    // check at_hash
+    if (access_token && jwtClaims.at_hash && !utils.checkHashValueRS256(access_token, jwtClaims.at_hash))
+      return self.fail("invalid at_hash");
+
+    // return jwt claims and jwt claims string
+    var idTokenSegments = id_token.split('.');
+    var jwtClaimsStr = new Buffer(idTokenSegments[1], 'base64').toString();
```
tushargupta51: This encoding should be base64url.

lib/aadutils.js
tushargupta51: The default encoding in update method is "utf8". We should hash the 'ascii' encoding of the input (code or access token).

tushargupta51: What if the value of 'content' is null? Will the update or digest method throw in that case? Also, createHash can throw if the OpenSSL version on platform does not support it: https://nodejs.org/api/crypto.html#crypto_crypto_createhash_algorithm, though that would be very unlikely  guess for sha256.

RandalliLama:
:shipit:

lib/nonceHandler.js
lovemaths: @tushargupta51 We are comparing nonce_provided and nonce_in_idToken, since nonce_provided is never null, there is no need to check nonce_in_idToken is null or not.

addev-ashish:
:shipit:

lib/nonceHandler.js
lovemaths: @tushargupta51 made the changes based on all your comments, see the latested commit

lib/oidcstrategy.js
```diff
@@ -25,24 +25,28 @@
...
-const Validator = require('./validator').Validator;
-const Log = require('./logging').getLogger;
 const InternalOAuthError = require('./errors/internaloautherror');
 const InternalOpenIDError = require('./errors/internalopeniderror');
+const Log = require('./logging').getLogger;
 const Metadata = require('./metadata').Metadata;
+const nonceHandler = require('./nonceHandler');
+const setup = require('./oidcsetup');
+const stateHandler = require('./stateHandler');
+const utils = require('./aadutils');
```
brentschmaltz: utils variable should be named 'aadutils'. This throws the ordering off.

lib/oidcstrategy.js
```diff
@@ -25,24 +25,28 @@
...
-const utils = require('./aadutils');
-const setup = require('./oidcsetup');
-const Validator = require('./validator').Validator;
-const Log = require('./logging').getLogger;
 const InternalOAuthError = require('./errors/internaloautherror');
 const InternalOpenIDError = require('./errors/internalopeniderror');
+const Log = require('./logging').getLogger;
 const Metadata = require('./metadata').Metadata;
+const nonceHandler = require('./nonceHandler');
+const setup = require('./oidcsetup');
```
brentschmaltz: 'setup' variable should be named 'oidcsetup'. This throws the ordering off.

lib/oidcstrategy.js
```diff
@@ -752,4 +577,494 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
 
+/**
+ * validate id_token, and pass the validated claims and the payload to callback
+ * if code (resp. access_token) is provided, we will validate the c_hash (resp at_hash) as well
+ *
+ * @param {String} id_token
+ * @param {Object} req
+ * @param {Function} callback
+ * @param {String} code (if you want to validate c_hash)
+ * @param {String} access_token (if you want o validate at_hash)
```
brentschmaltz: These should be in the same order as the function prototype.

TYPO on access_token (if you want o -> is you want to)

lib/oidcstrategy.js
```diff
@@ -159,6 +203,10 @@ function Strategy(options, verify) {
   // validator will throw exception if a required option is missing
   var validator = new Validator(validatorConfiguration);
   validator.validate(itemsToValidate);
+
```
brentschmaltz: Please have consistent line breaks after clauses. For example after ';' when validatorConfiguration is set, put a blank line.

brentschmaltz: Added issue for allowing http by default: https://github.com/AzureAD/passport-azure-ad/issues/164

lib/oidcstrategy.js
```diff
@@ -207,350 +299,81 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
+       *     This case we get a 'code', we will use it to get 'access_token' and 'id_token'
+       * (5) for any other request, we will ask for authorization and initialize the authorization process 
+       */
+      (next) => {
+        var err, err_description, id_token, code;
+        err = err_description = id_token = code = null;
+
+        if (req.query && req.query.error) err = req.query.error;
+        if (req.query && req.query.error_description) err_description = req.query.error_description;
+        if (req.query && req.query.id_token) id_token = req.query.id_token;
```
brentschmaltz: we should not accept id_token from a query.

lib/oidcstrategy.js
```diff
@@ -207,350 +299,81 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
+        if (req.query && req.query.code) code = req.query.code;
+
+        if (req.body && req.body.error) err = req.body.error;
+        if (req.body && req.body.error_description) err_description = req.body.error_description;
+        if (req.body && req.body.id_token) id_token = req.body.id_token;
+        if (req.body && req.body.code) code = req.body.code;
```
brentschmaltz: We need to make sure the 'id_token' and 'code' came from the POST. this code will allow the 'code' in the query and 'id_token' in the post

lib/oidcstrategy.js
```diff
@@ -207,350 +299,81 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
+
+        if (!err && !id_token && !code) {
+          // ask for authorization, initialize the authorization process
+          return self._flowInitializationHandler(req, next);
```
brentschmaltz: is this code expected to handle 'code' only flows?

lib/oidcstrategy.js
brentschmaltz: Why are we hard coding RS256. Use the 'alg' value to determine the HASH algorithm to use.

lib/oidcstrategy.js
lovemaths: _flowInitializationHandler is used to do the redirection for authentication

lib/oidcstrategy.js
```diff
@@ -752,4 +577,494 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+
+  // decode id_token
+  const decoded = jws.decode(id_token);
+  if (decoded == null)
+    return self. fail(null, false, 'Invalid JWT token');
+  log.info('token decoded: ', decoded);
+
+  // get Pem Key
+  var PEMkey = null;
+  if (decoded.header.x5t) {
```
brentschmaltz: I would reorder to check 'kid' first then 'x5t' as 'x5t' is an artifact of an older world and most likely will be dropped.

lib/oidcstrategy.js
```diff
@@ -645,7 +646,7 @@ Strategy.prototype._validateIdToken = function validateIdToken(id_token, code, a
 
     // return jwt claims and jwt claims string
     var idTokenSegments = id_token.split('.');
-    var jwtClaimsStr = new Buffer(idTokenSegments[1], 'base64').toString();
+    var jwtClaimsStr = base64url(new Buffer(idTokenSegments[1], 'base64'));
```
tushargupta51: @lovemaths I think you should use base64url.decode. https://github.com/brianloveswords/base64url#base64urldecodeinput-string-encoding-string--utf8-string

What we want here is to decode the base64url encoded payload into string. Use something like this:
`base64url.decode(idTokenSegments[1])`
This will return the string in utf8 encoding.

lib/oidcstrategy.js
```diff
@@ -207,350 +299,81 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
+      (next) => {
+        var err, err_description, id_token, code;
+        err = err_description = id_token = code = null;
+
+        if (req.query && req.query.error) err = req.query.error;
+        if (req.query && req.query.error_description) err_description = req.query.error_description;
+        if (req.query && req.query.id_token) id_token = req.query.id_token;
+        if (req.query && req.query.code) code = req.query.code;
+
+        if (req.body && req.body.error) err = req.body.error;
```
brentschmaltz: error has three parts, we are missing error_uri
http://openid.net/specs/openid-connect-core-1_0.html#AuthError

opened: https://github.com/AzureAD/passport-azure-ad/issues/166

lib/oidcstrategy.js
```diff
@@ -866,7 +880,13 @@ Strategy.prototype._authCodeFlowHandler = function authCodeFlowHandler(code, req
 
         log.info('received id_token: ', id_token);
 
-        return self._validateIdToken(id_token, null, access_token, req, (jwtClaimsStr, jwtClaims) => {
+        return self._validateResponse(id_token, null, access_token, req, (jwtClaimsStr, jwtClaims) => {
+          // for 'code id_token' flow, check iss/sub from the first id_token with those from this id_token
+          if (issFromPrevIdToken && issFromPrevIdToken !== jwtClaims.iss)
+            return self.fail('iss from the second id_token does not match iss from the first id_token');
```
tushargupta51: Instead of using first and second, you can use something like:
iss in the id_token from the token endpoint does not match the iss in the id_token from the authorization endpoint.

lib/oidcstrategy.js
```diff
@@ -866,7 +880,13 @@ Strategy.prototype._authCodeFlowHandler = function authCodeFlowHandler(code, req
...
+          if (subFromPrevIdToken && subFromPrevIdToken !== jwtClaims.sub)
+            return self.fail('sub from the second id_token does not match sub from the first id_token');
```
tushargupta51: same remark as above.

lib/oidcstrategy.js
lovemaths: ok, will change it.

lib/oidcstrategy.js
brentschmaltz: It is not clear to me what error the user will receive if the JWT signature is empty. By default we should require a signature on the tokens.

lib/oidcstrategy.js
```diff
@@ -187,10 +273,16 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
   async.waterfall(
     [
-      // compute metadata url
+      /* 
+       * Step 1. compute metadata url 
+       */
       (next) => {
         // B2C interception
-        let metadata = self._options.identityMetadata;
+        let metadataUrl = self._options.identityMetadata;
```
jeffwilcox: Any particular reason for using `let` so often vs moving straight to `const`?

lib/oidcstrategy.js
```diff
@@ -207,350 +299,81 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
-          metadata = self._options.identityMetadata
+          metadataUrl = self._options.identityMetadata
             .replace('common', self._options.tenantName)
             .concat(`?p=${policy}`);
 
-          self._cacheKey = policy; // this policy will become cache key.
-
-          log.info('B2C: New Metadata url provided to Strategy was: ', metadata);
-        }
+          cachekey = 'policy: ' + policy; // this policy will become cache key.
```
jeffwilcox: On our team we've been moving to using the newer template strings in ES6, so this would be:

```
cacheKey = `policy: ${policy}`'
```

Just something to consider but just a nit.

lib/oidcstrategy.js
```diff
@@ -579,21 +402,20 @@ Strategy.prototype.setOptions = function setOptions(options, metadataUrl, done)
...
         });
       }, { ttl }, next);
     },
     // merge fetched metadata with options
     function loadOptions(metadata, next) {
-      // fetched metadata always takes precendence over configured options
+      self.metadata = metadata;
 
-      self._skipUserProfile = !!options.skipUserProfile;
+      // fetched metadata always takes precendence over configured options
```
jeffwilcox: spelling - `precedence`

lib/oidcstrategy.js
lovemaths: line 74 of index.js in jsonwebtoken:

```
  var hasSignature = parts[2].trim() !== '';

  if (!hasSignature && secretOrPublicKey){
    return done(new JsonWebTokenError('jwt signature is required'));
  }
```
the error raised will go into the 'err' in the callback in our code:
```
return jwt.verify(id_token, PEMkey, options, (err, jwtClaims) => {
```

lib/oidcstrategy.js
lovemaths: ok

lib/oidcstrategy.js
lovemaths: ok

lib/oidcstrategy.js
lovemaths: ok

lib/oidcstrategy.js
lovemaths: validatorConfiguration and the validate are closely related, I think they should stay together without the line break.

lib/oidcstrategy.js
```diff
@@ -752,4 +577,494 @@ Strategy.prototype._shouldLoadUserProfile = function shouldLoadUserProfile(issue
...
+    if (Array.isArray(scope)) {
+      scope = scope.join(config.scopeSeparator);
+    }
+    if (scope) {
+      params.scope = ['openid', scope].join(config.scopeSeparator);
+    } else {
+      params.scope = 'openid';
+    }
+
+    // if (policy) { params['p'] = policy; }; // Policy parameter should be included.
```
jeffwilcox: Not sure what your team's practices are, but should this have a `TODO:` or `MUSTFIX:` prefix in a comment, or anything, or is this commented out long-term by design?

lib/stateHandler.js
jeffwilcox: Spacing feels different on this file.

Is there an `eslint` or `jslint` config setup that can be used for this project to keep things consistent?

Looks like 4-space not 2-space here...

lib/oidcstrategy.js
lovemaths: ok

lib/oidcstrategy.js
lovemaths: ok

lib/oidcstrategy.js
lovemaths: ok

lib/oidcstrategy.js
lovemaths: ok

lib/oidcstrategy.js
lovemaths: the comment was there for a long time, I will change it to a TODO

lib/stateHandler.js
lovemaths: will edit the space

lib/oidcstrategy.js
lovemaths: yes

jeffwilcox:
Thanks... should I consider moving my dev deployment/env to the `dev` branch now to help bake some of this?

jeffwilcox:
... ah, nevermind. I saw it get closed but didn't realize it isn't merged.

Issue #168, opened 2016-08-16
The latest NPM module has spacing problems around certificates


richardspence:
Install the latest 
    npm install passport-azure-ad --save

In AADUtil.js the spacing is:
```
 const pem = `-----BEGIN RSA PUBLIC KEY-----\n
    ${derB64.match(/.{1,64}/g).join('\n')}
    \n-----END RSA PUBLIC KEY-----\n`;

  return pem;
```
In master it is 
```
  const derB64 = new Buffer(encodedPubkey, 'hex').toString('base64');

  let pem = `-----BEGIN RSA PUBLIC KEY-----\n`;
  pem += `${derB64.match(/.{1,64}/g).join('\n')}`
  pem += `\n-----END RSA PUBLIC KEY-----\n`;

  return pem;
};
```

The npm package has invalid spacing, since it will encode the tabs and extra linebreaks.  Can you publish a new NPM package?

lovemaths:
@richardspence We published a new npm package yesterday, please check it out.

PR #169, opened 2016-08-16
Fixed an invalid conditional

richardspence:
The conditional was to short circuit key checking but it never worked as intended

### DIFF
--- a/lib/metadata.js
+++ b/lib/metadata.js
@@ -183,7 +183,7 @@ Metadata.prototype.generateOidcPEM = function generateOidcPEM(kid) {
     log.info('working on key:', key);
 
     // are we working on the right key?
-    if (!key.kid === kid) {
+    if (key.kid !== kid) {
       return false;
     }
 


polita:
@richardspence Thanks for your contribution. This is already fixed in the dev branch. If you're contributing or pulling current sources, please use the dev branch.

richardspence:
Ok good to know.  In the current NPM package, this is not fixed.  I also can't used npm install {this git url} as a workaround.  Can you publish these changes to the NPM package, because right now the BearerStrategy is broken because of it?

richardspence:
Also related #168

polita:
We're working on a release in the next week or so. It'll be on npm as well.

richardspence:
Ok sounds good.  I'm unblocked by using my own fork, but I just want to be clear, I find the NPM package as completely unusable in it's current state (note: I'm only trying to use the bearerstrategy).  You should consider expediting the release.

PR #180, opened 2016-08-24
remove jwt dependency and tokenValidator.js

lovemaths:
(1) write our own jwt for token validation
(2) remove tokenValidator.js since it is not used anywhere


lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+exports.verify = function(jwtString, PMEKey, options, callback) {
+
+  /*********************************************************************
+   * Checking parameters
+   ********************************************************************/
+
+  // All of the four parameters are required
+  if (!jwtString)
+    throw new Error('jwtString must be provided in jsonWebToken.verify');
+  if (!PMEKey)
```
polita: typo on PEMKey

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+ * @callback
+ */
+exports.verify = function(jwtString, PMEKey, options, callback) {
+
+  /*********************************************************************
+   * Checking parameters
+   ********************************************************************/
+
+  // All of the four parameters are required
+  if (!jwtString)
```
polita: Should we check against empty string?

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+    var args = Array.prototype.slice.call(arguments, 0);
+    return process.nextTick(function() {
+      callback.apply(null, args);
+    });
+  };
+
+  // make sure we have the required fields in options
+  if (!options.audience)
+    return done(new Error('options.audience is missing in jsonWebToken.verify'));
+  if (!options.algorithms)
```
polita: Is options.algorithms an array? If so, do we need to check for length > 0? How about "none"?

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+    return done(new Error('missing payload in the token'));
+  if (!signature)
+    return done(new Error('missing signature in the token'));
+
+  /*********************************************************************
+   * validate algorithm and signature
+   ********************************************************************/
+
+  // alg in the token header should be one of those provided in options.algorithms
+  if (!~options.algorithms.indexOf(header.alg)) {
```
polita: please change this to 
`if (options.algorithms.indexOf(header.alg) === -1) {`
The tilde is tricky, but it's also hard to read.

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+    return done(new Error('invalid algorithm'));
```
polita: Should we be checking for header.alg == "none"?

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+   * validate payload content
+   ********************************************************************/
+
+  // (1) issuer
+  //   - check the existence and the format of payload.iss
+  //   - validate if options.issuer is set
+  if (typeof payload.iss !== 'string')
+    return done(new Error('invalid iss value in payload'));
+  if (options.issuer && options.issuer !== payload.iss)
+    return done(new Error('jwt issuer is invalid. expected: ' + options.issuer));
```
polita: Are we allowing options.issuer and payload.iss to both equal ""? We should throw if no issuer is specified.

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+
+  // (2) subject
+  //   - check the existence and the format of payload.sub
+  //   - validate if options.subject is set
+  if (typeof payload.sub !== 'string')
+    return done(new Error('invalid sub value in payload'));
+  if (options.subject && options.subject !== payload.sub)
```
polita: Check for subject not empty string?

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+      return done(new Error('jwt subject is invalid. expected: ' + options.subject));
+
+  // (3) audience
+  //   - always validate
+  //   - allow payload.aud to be an array of audience
+  //   - options.audience must be a string
+  if (typeof options.audience !== 'string')
+    return done(new Error('invalid options.audience value'));
+  var payload_audience = Array.isArray(payload.aud) ? payload.aud : [payload.aud];
+  if (payload_audience.indexOf(options.audience) == -1)
```
polita: Check for empty string in both

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+  if (typeof payload.exp !== 'number')
+    return done(new Error('invalid exp value in payload'));
+  if (!options.ignoreExpiration) {
+    if (Math.floor(Date.now() / 1000) >= payload.exp) {
+      return done(new Error('jwt is expired'));
+    }
+  }
+
+  // (5) iat
+  //   - check the existence and the format of payload.exp
```
polita: heck the existence and the format of payload.*iat*

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+  if (typeof payload.iat !== 'number')
```
polita: Will this fail if iat isn't present? It's not required.

lib/jsonWebToken.js
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  /*********************************************************************
+   * validate payload content
+   ********************************************************************/
+
+  // (1) issuer
+  //   - check the existence and the format of payload.iss
+  //   - validate if options.issuer is set
+  if (typeof payload.iss !== 'string')
```
polita: This code will need to handle multiple issuers being handed in to support issuer validation on the common endpoint. options.issuer may need to be able to be an array.

lib/jsonWebToken.js
lovemaths: iat is required in id_token according to the spec, we only check the existence since we don't use it.

(http://openid.net/specs/openid-connect-core-1_0.html#IDToken)

iat
REQUIRED. Time at which the JWT was issued. Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time.

polita: Ok, looks good.

lib/jsonWebToken.js
lovemaths: done

lib/jsonWebToken.js
lovemaths: we fail when we check if jwtString has 3 parts. Anyway, I just added the check here.

lib/jsonWebToken.js
lovemaths: done

lib/jsonWebToken.js
lovemaths: done

lib/jsonWebToken.js
lovemaths: done

lib/jsonWebToken.js
lovemaths: enough to check if payload.iss is "" since later we will compare them. Now added the check

lib/jsonWebToken.js
lovemaths: done

lib/jsonWebToken.js
lovemaths: added check for options.audience, since later we compare with payload.aud, this will suffice.

lib/jsonWebToken.js
lovemaths: done

lib/jsonWebToken.js
lovemaths: done.

lib/jsonWebToken.js
```diff
@@ -0,0 +1,195 @@
...
+   * validate payload content
+   ********************************************************************/
+
+  // (1) issuer
+  //   - check the existence and the format of payload.iss
+  //   - validate if options.issuer is set
+  if (typeof payload.iss !== 'string' || payload.iss === '')
+    return done(new Error('invalid iss value in payload'));
+  if (options.issuer) {
+    if ((Array.isArray(options.issuer) && options.issuer.indexOf(payload.iss) == -1) ||
```
polita: @brentschmaltz Can you comment on this? To support multiple issuers, we are support options.issuer as an Array, but would payload.iss come back as an array of issuers as well if we're going against the common metadata endpoint?
@lovemaths It might be worth just trying it. What does payload.iss return when you go against common?

lovemaths: @polita @brentschmaltz  payload.iss is always an URL string starting with 'https', according to the spec (http://openid.net/specs/openid-connect-core-1_0.html#IDToken)

```
iss
REQUIRED. Issuer Identifier for the Issuer of the response. The iss value is a case sensitive URL using the https scheme that contains scheme, host, and optionally, port number and path components and no query or fragment components.
```

I tried the common endpoint as well, the issuer is always a string like "https://sts.window.net/....".

polita: @lovemaths Can you chat with @brentschmaltz or @tushargupta51 about how we handle extensibility on issuer to handle the common endpoint for Wilson? We should ensure the design is similar.

lovemaths:
@polita talked with @tushargupta51 , the design is the same, except Wilson allows developer to provide their own function to validate issuer, I don't plan to have such a thing for passport though.

polita:
Sounds good, thanks. Please get @brentschmaltz to look at it. :shipit: from me, but get one from him or @tushargupta51 too.

lib/jsonWebToken.js
```diff
@@ -0,0 +1,195 @@
...
+    });
+  };
+
+  // make sure we have the required fields in options
+  if (typeof options.audience !== 'string' || options.audience === '')
+    return done(new Error('options.audience is missing in jsonWebToken.verify'));
+  if (!options.algorithms)
+    return done(new Error('options.algorithms is missing in jsonWebToken.verify'));
+  if (!Array.isArray(options.algorithms) || options.algorithms.length == 0 ||
+    (options.algorithms.length == 1 && options.algorithms[0] === 'none'))
```
tushargupta51: Why using "==" for comparison here?

lib/jsonWebToken.js
```diff
@@ -0,0 +1,195 @@
...
+    return done(new Error('jwtString is malformed'));
+  if (parts[2] === '')
+    return done(new Error('signature is missing in jwtString'));
+  
+  // decode jwsString
+  var decodedToken;
+  try {
+    decodedToken = jws.decode(jwtString);
+  } catch(err) {
+    return done(new Error('invalid token'));
```
tushargupta51: Maybe we should have more information in the error here like decoding the token failed.

lib/jsonWebToken.js
lovemaths: when it comes here, options.algorithms is an array, and .length always gives a number, "==" and "===" are the same.

lib/jsonWebToken.js
lovemaths: Changed it to "failed to decode the token"

lib/jsonWebToken.js
```diff
@@ -0,0 +1,195 @@
...
+
+  // split jwtString, make sure we have three parts and we have signature
+  var parts = jwtString.split('.');
+  if (parts.length !== 3)
+    return done(new Error('jwtString is malformed'));
+  if (parts[2] === '')
+    return done(new Error('signature is missing in jwtString'));
+  
+  // decode jwsString
+  var decodedToken;
```
tushargupta51: We are decoding token twice. Once in validateResponse in oidcStrategy.cs and once here. Should we think about passing decoded token to verify method?

lib/jsonWebToken.js
```diff
@@ -0,0 +1,195 @@
...
+  if (options.subject && options.subject !== payload.sub)
+      return done(new Error('jwt subject is invalid. expected: ' + options.subject));
+
+  // (3) audience
+  //   - always validate
+  //   - allow payload.aud to be an array of audience
+  //   - options.audience must be a string
+  if (typeof options.audience !== 'string' || options.audience === '')
+    return done(new Error('invalid options.audience value'));
+  var payload_audience = Array.isArray(payload.aud) ? payload.aud : [payload.aud];
```
tushargupta51: We do not allow multiple audiences in the payload and throw and error in oidcStrategy here: https://github.com/AzureAD/passport-azure-ad/blob/sijliu/v3.0.0dev/lib/oidcstrategy.js#L734

I think we should move that check before jwt.Verify is called or throw here when payload.aud is Array.

lib/jsonWebToken.js
tushargupta51: "==" does type conversion, whereas "===" does not. We use both in the code. I think we should use one consistently.

lib/jsonWebToken.js
tushargupta51: @lovemaths What happens when options.validateIssuer is false but options.issuer is not null?

lib/jsonWebToken.js
lovemaths: As I said, since .length always gives a number, there will be no type conversion going on, "==" and "===" are the same. I can change it to "===".

lib/jsonWebToken.js
lovemaths: @tushargupta51 later at line 125 we need to use the original string to validate signature.

lib/jsonWebToken.js
lovemaths: @tushargupta51 done

lib/jsonWebToken.js
lovemaths: @tushargupta51 We ignore options.issuer. See the new commit.

lib/jsonWebToken.js
```diff
@@ -32,7 +32,9 @@ const jws = require('jws');
  * @options
  *   - algorithms (required)
  *   - audience (required)
- *   - issuer (optional, validate if provided)
+ *   - allowMultiAudiencesInToken (optional, default is true)
```
tushargupta51: where are we setting the default for this property?

tushargupta51:
:shipit:

PR #182, opened 2016-08-26
remove dead code and simplify oidcstrategy

lovemaths:
(1) remove discovery folder, they are dead code
(2) OIDCStrategy.prototype.configure is used to pick parameters required for oauth flow (to get access_token, useinfo etc using OAuth2 library), and save these parameters in some object called 'config', so later on OAuth2 can use things in 'config'. The current code makes it way too complicated. We delete oidcsetup.js and oidcconfig.js, and remove OIDCStrategy.prototype.configure. We add a simple function
OIDCStrategy.prototype.configOauth, which does the parameter picking job.

Commit 1 is for deleting stuff, and commit 2 is for formatting the code, this way it's easier to see what changed.


lib/oidcstrategy.js
```diff
@@ -863,147 +863,147 @@ Strategy.prototype._authCodeFlowHandler = function authCodeFlowHandler(code, req
...
+        const sub = jwtClaims.sub;
+        const iss = jwtClaims.iss;
 
-          return self._shouldLoadUserProfile(iss, sub, (shouldLoadUserProfileError, load) => {
-            if (shouldLoadUserProfileError) {
-              return next(shouldLoadUserProfileError);
+        return self._shouldLoadUserProfile(iss, sub, (shouldLoadUserProfileError, load) => {
+          if (shouldLoadUserProfileError) {
+            return next(shouldLoadUserProfileError);
+          }
```
lovemaths: This commit I only shift those code to left by a tab, the diff looks kind of weird, maybe because of mixed use of tab and blanks.

polita: Please replace all tabs with spaces. It's easy to do in VS or Notepad++.
http://stackoverflow.com/questions/455037/convert-tabs-to-spaces-in-notepad

lovemaths: I have two other pending PR's on oidcstrategy, I cannot replace the tabs with spaces for this PR since that will cause a big merge conflict. I removed the second commit, I will use a separated PR for the tab replacement after all of these 3 PR's in.

lovemaths: Re-did the second commit using blank, nothing changes. Seems to be git diff problem.

lib/oidcstrategy.js
```diff
@@ -45,7 +45,6 @@ const InternalOpenIDError = require('./errors/internalopeniderror');
 const Log = require('./logging').getLogger;
```
polita: Please clean up the capitalization for these consts. They should all be lower case or all be upper case.

polita:
Pending the one comment, :shipit:

lovemaths:
@polita We use upper case for class name, like "Metadata" and "Log" ("self.metata = new Metadata(...)",  "const log = new Log(...) "); anything else is an instance, we use lower case. The only thing that violates this is "Validator", I will change it to lower case.

lovemaths:
@polita Just find out "Validator" is also a class name (var validator = new Validator(validatorConfiguration);), so no need to change it as well.

polita:
This is super confusing when you read the code. You can't tell what's a const type and what's a const instance. Can you group types and instances, and put a comment for types and a comment for instances?

lovemaths:
@polita Sure, I will do it after the jwtWebToken PR is in so I don't need to resolve conflicts.

polita:
Ok, sounds good. :shipit:

PR #184, opened 2016-08-30
remove _shouldLoadUserProfile

lovemaths:
This function assumes options.skipUserProfile can be a function, this is impossible since options.skipUserProfile is read from the config file, it can only be true or false. So let's remove this dead code.

For any code like:

```
self._shouldLoadUserProfile(iss, sub, (err, load) => {
     if (err)
        err_handling_logic;
     callback_function_body;
   }
);
```

We don't need _shouldLoadUserProfile to decide if we want to load user profile, so remove the _shouldLoadUserProfile and transform it to
```
 // we don't load userinfo unless user set skipUserProfile to false
const load = (self._options.skipUserProfile === false);

callback_function_body;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,15 @@
-<a name="2.0.1"></a>
+<a name="3.0.0"></a>
+# 3.0.0
+
+## Breaking changes from 2.0.1
+
+### OIDCStrategy
+* skipUserProfile option is no longer provided. We will load 'userinfo' if AAD v1 is being used and
+if there is 'code' involved in the flow (in other words, one of the following flows: 'code', 
+'code id_token', 'id_token code'). For all other scenarios, we do an 'id_token' fallback, since we 
+cannot obtain 'userinfo' (AAD v2 doesn't have an userinfo endpoint, and 'id_token' flow doesn't yield
+ an 'access_token' to exchange for an 'userinfo' token).
+
 # 2.0.1
 
 ## Major changes from 2.0.0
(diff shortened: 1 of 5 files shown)


polita:
@brandwe What was the intention behind options.skipUserProfile supporting a function? Was that always dead code?

lovemaths:
@polita That is intended to allow developer to provide their own logic of deciding if load userinfo or not. This is unnecessary, they just need to provide 'true' or 'false' for that field.

brandwe:
It's slightly more complicated. Passport.js provides for the default User Info Endpoint flow where we grab data from that place and then populate a profile object. We don't have User Info Endpoint (yet) so instead I populate the profile object with data from the ID Token. 

However eventually we will have a User Info Endpoint so for now I kept the code in there from the forked library but pass a way to "skipUserProfile"

polita:
Why would anyone want to skip loading the user profile? What's the scenario?

lovemaths:
@brandwe I think v1 has userinfo endpoint but v2 doesn't.

brandwe:
Indeed. The scenario is if the IdP doesn't have a User Info Endpoint (like us)

lovemaths:
@brandwe @polita Our treatment of userinfo is that 
(1) we load userinfo if user set skipUserProfile false and it's v1
(2) we use id_token otherwise
(3) if user sets skipUserProfile true but it's v2, we give a warning

skipUserProfile only needs to be true or false for us, and we don't care about any other IdP's.

Please either reject this PR with a good reason or give a shipit. Thanks.

polita:
If the only reason to skip loading the user profile is that you're in V2 where there's no endpoint, should we error if you set it to true in V1 and if you set it to false in V2? Given that it's completely dependent on whether there's a user info endpoint, why have this as a user setting at all?

lovemaths:
@polita That makes sense, we should remove skipUserProfile from the setting, we always load userinfo from v1, and always don't load from v2. I will make the change.

polita:
@lovemaths Thanks Sijun.

lovemaths:
@polita Removed skipUserProfile option in the second commit.

polita:
:shipit:

Issue #185, opened 2016-08-30
Bug: in BearerStrategy: policyName is not pulled from options


brandwe:
`        // use the policy name as the metadata cache key
        cacheKey = 'policy: ' + policyName;`

This fails. Should:

-  use the policyName in options.

PR #186, opened 2016-08-30
Fix for issue #185 - BearerStrategy profileName issue.

brandwe:


lib/bearerstrategy.js
```diff
@@ -293,8 +293,9 @@ Strategy.prototype.authenticate = function authenticateStrategy(req) {
 
       var metadataURL = self._options.identityMetadata;
 
-      if (self._options.policyName) {
-        log.info('B2C: using policy %s', self._options.policyName);
+      if (!self._options.policyName) {
```
polita: Should we really be erroring any time there's no policyName? What if we're not using B2C- there'd be no policy here, right? And why remove the log info on the policy?

brandwe: You are right, but we have a bigger problem. I had two code paths - one for B2C and one for v2. It seems that this has been removed. A lot is now broken if that is the case, because this code now:

* Automatically does changes to the metadata endpoint (removes common)
* Automatically uses policyName when evaluating policies.

We need to have two forks in the waterfalls like I had in the old code base. I'm open to other options as well.

Right now this is effectively a B2C BearerStrategy. 🎱

brandwe: So what I realize I did when I first added B2C was use policyName as the trigger for doing B2C work, so logging that it isn't there isn't correct behavior. Reverted.

lib/bearerstrategy.js
```diff
@@ -294,7 +294,7 @@ Strategy.prototype.authenticate = function authenticateStrategy(req) {
       var metadataURL = self._options.identityMetadata;
 
       if (self._options.policyName) {
-        log.info('B2C: using policy %s', self._options.policyName);
+      log.info('B2C: using policy %s', self._options.policyName);
```
polita: Why does the indentation look off here? Where are we closing the open bracket on the if line above?

brandwe: As I said before, this is the signal to use the B2C flow, so this is if statement is quite large (it sets the stage for B2C, including creating a new caching key for policy). The end of the if statement is line number 312: https://github.com/AzureAD/passport-azure-ad/blob/e7305c2cfc7f6053fec9d8cbc5797791fff5c7f6/lib/bearerstrategy.js#L312

lovemaths: @brandwe  There should be two extra blanks in front of 297.

polita: @brandwe You should just revert the change on line 297, as @lovemaths says.

polita:
@lovemaths Please test this against a B2C tenant before we merge the change. Looks like we might need a fix to #187 before the E2E will work.

PR #197, opened 2016-09-09
change size

lovemaths:


polita:
:shipit: Please also make changes to versions in preparation for the release.

PR #198, opened 2016-09-09
size change

lovemaths:


polita:
This is for 2.0.2, right? Please make a branch for 2.0.2 so we can update the version and other details in preparation for a release. We'll move it all into master when we create the release.

lovemaths:
Call the branch release2x?

PR #202, opened 2016-09-10
update changelog and readme

lovemaths:
(1) added changes in 1.x to changelog of 2.x
(2) update the minimum recommended version

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ _passport-azure-ad_ has a known security vulnerability affecting versions <1.4.6
 
 ## Versions
 Current version - 2.0.2  
-Minimum  recommended version - 1.4.7  
+Minimum  recommended version - 1.4.6  
 You can find the changes for each version in the [change log](https://github.com/AzureAD/passport-azure-ad/blob/master/CHANGELOG.md).
 
 ## Contribution History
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -17,7 +17,7 @@ _passport-azure-ad_ has a known security vulnerability affecting versions <1.4.6
 
 ## Versions
 Current version - 2.0.2  
-Minimum  recommended version - 1.4.6  
+Minimum  recommended version - 1.4.7
```
polita: Per other PR, please leave this at 1.4.6. Sorry for the confusion.

lovemaths: changed back to 1.4.6

lib/bearerstrategy.js
```diff
@@ -308,7 +308,7 @@ Strategy.prototype.authenticate = function authenticateStrategy(req) {
             .concat(`&${aadutils.getLibraryVersionParameterName()}=${aadutils.getLibraryVersion()}`);
 
         // use the policy name as the metadata cache key
-        cacheKey = 'policy: ' + policyName;
+        cacheKey = 'policy: ' + self._options.policyName;
```
polita: Do you mean to include this change? B2C will still be broken even with this change, right?

lovemaths: Yes. B2C works for BearerStrategy, since we only consume the token and the policy is given in the config. I tested the change with sample. For OIDCStrategy we need to figure out the policy.

PR #216, opened 2016-09-16
multiple state and nonce support

lovemaths:
We used to save only one state in session. Now
(1) we save an array of {state: state_value, timeStamp: time_stamp_value} in
session, where timestamp is in increasing order.
(2) user can customize the array max size and the lifetime of state, the
default max size is 200 and the default lifetime is 30 minutes.
(3) when we add new state or validate state, we remove the expired state, and
we might remove the oldest several states if the size exceeds the max size.

We do the almost same thing to nonce. The only difference is: when we verify
a state, sessionContentHandler.verify can extract state from request if state
is not provided; when we verify nonce, we have to provided the nonce to the
sessionContentHandler.verify function.

Due to the similarity of handling nonce and state in session, we removed the
stateHandler.js and nonceHandler.js, now we use sessionContentHandler for
both of them.

Also reordered the packages at the beginning of oidcstrategy.js, removed
comments about 'realm' since it is not used anywhere in the code.


lib/oidcstrategy.js
```diff
@@ -144,20 +146,26 @@ function onProfileLoaded(strategy, args) {
...
  *   identityMetadata: config.creds.identityMetadata,
  *   responseType: config.creds.responseType,
  *   responseMode: config.creds.responseMode,
  *   validateIssuer: config.creds.validateIssuer,
  *   passReqToCallback: config.creds.passReqToCallback,
- *   loggingLevel: config.creds.loggingLevel
+ *   loggingLevel: config.creds.loggingLevel,
+ *   stateMaxAmount: config.creds.stateMaxAmount,
+ *   stateMaxAge: config.creds.stateMaxAge,
+ *   nonceMaxAmount: config.creds.nonceMaxAmount,
```
polita: I can understand why someone would want to configure the state and nonce max age, but they would do so as one setting, right? Why would they, for instance, set state to 1 hour but nonce to 5 hours? How do we let you configure this in Wilson?

lib/oidcstrategy.js
```diff
@@ -144,20 +146,26 @@ function onProfileLoaded(strategy, args) {
...
  *   scope: config.creds.scopes,
  *   clientID: config.creds.clientID,
  *   identityMetadata: config.creds.identityMetadata,
  *   responseType: config.creds.responseType,
  *   responseMode: config.creds.responseMode,
  *   validateIssuer: config.creds.validateIssuer,
  *   passReqToCallback: config.creds.passReqToCallback,
- *   loggingLevel: config.creds.loggingLevel
+ *   loggingLevel: config.creds.loggingLevel,
+ *   stateMaxAmount: config.creds.stateMaxAmount,
```
polita: Do we let you configure the stateMaxAmount in Wilson? I have no idea what I'd set it to.

lib/sessionContentHandler.js
```diff
@@ -0,0 +1,98 @@
...
+
+/*
+ * the handler for state/nonce
+ * @identifier        - such as 'nonce' or 'state'
+ * @maxAmout          - the max amout of 'identifier' you want to save in the session, default is 200
+ * @maxAge            - when the 'identifier' in session expires in seconds, default is 1800 (seconds)
+ */ 
+function SessionContentHandler(identifier, maxAmount, maxAge) {
+  if (!identifier || identifier === '')
+    throw new Error('identifier is required to use sessionContentHandler');
```
polita: 'SessionContentHandler: identifier is a required parameter in sessionContentHandler'

lib/sessionContentHandler.js
```diff
@@ -0,0 +1,98 @@
...
+  if (maxAmount && (typeof maxAmount !== 'number' || maxAmount <= 0 || maxAmount % 1 !== 0))
+    throw new Error('SessionContentHandler: maxAmount must be a positive integer');
+  if (maxAge && (typeof maxAge !== 'number' || maxAge <= 0))
+    throw new Error('SessionContentHandler: maxAge must be a positive number');
+  this.identifier = identifier;
+  this.maxAge = maxAge || 1800;  // seconds
```
polita: Are we setting default values here? The default values shoudl be set on the config properties, not here. Just error if it's not a sane value here.

lib/sessionContentHandler.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  // the array in session
+  var array = req.session[sessionKey] && req.session[sessionKey][this.identifier];
+  if (!array)
+    array = [];
+
+  // remove the expired ones in array, and check if we have the itemToVerify in array
+  aadutils.processArray(array, this.maxAmount, this.maxAge);
+  var valid = aadutils.findAndDeleteInArray(array, this.identifier, itemToVerify);
```
polita: Why are you deleting it from the array?

lib/sessionContentHandler.js
```diff
@@ -0,0 +1,98 @@
...
+
+  // clear emptry array, and clear the session if there is nothing inside
```
polita: Nit: "empty"

lib/sessionContentHandler.js
```diff
@@ -0,0 +1,98 @@
...
+  if (req.session[sessionKey] && array.length === 0)
+    delete req.session[sessionKey][this.identifier];
+  if (req.session[sessionKey] && Object.keys(req.session[sessionKey]).length ===0)
```
polita: Nit: space before 0.

lovemaths:
state and nonce are treated almost the same way, so below I will only talk about state.

We clear out expired states when we add a new state, and clear out the expired and the verified state when we verify a state, not when we reach max count. In other words, we clear out the expired and verified states for every request coming to passport.authenticate.

lib/sessionContentHandler.js
lovemaths: we delete verified state and nonce.

lib/oidcstrategy.js
lovemaths: Wilson doesn't use session.

lib/oidcstrategy.js
brentschmaltz: nonce configurations:
1. RequireTimeStampInNonce (bool - default true)
2. NonceLifetime (TimeSpan - default 60 minutes)

I don't also don't see the reason for a separate expiration value. If we view <state, nonce, policy> as a tuple, we can have one expiration.

brentschmaltz:
We had that conversation and landed on all oidc request artifacts (nonce, state, ...) for a session should be deleted once the oidc response has been successfully processed. Each request will have a unique set of artifacts stored as a tuple. Passport is responsible for assigning a unique session identifier.

lib/sessionContentHandler.js
polita: Got it, makes sense.

lib/oidcstrategy.js
lovemaths: User can configure the maxAge in Wilson, in my second commit, I change the default to 60 minutes to match Wilson. Wilson saves things in cookie, but passport saves them in session. Talked with Brent, we decided to set maxAmount to 10.

lib/oidcstrategy.js
lovemaths: In the second commit I changed them to a tuple, so now they have the same expiration.

lib/sessionContentHandler.js
lovemaths: identifier is no longer used. Removed this part.

lib/sessionContentHandler.js
lovemaths: Now set the default in the constructor of strategy and error it here.

lib/sessionContentHandler.js
lovemaths: done

lib/sessionContentHandler.js
lovemaths: done

lib/oidcstrategy.js
polita: Why do we need a maxAmount? Why not have only maxAge? Also, if Wilson calls it NonceLifetime, why don't we call it the same thing?

lib/oidcstrategy.js
```diff
@@ -198,8 +196,7 @@ function Strategy(options, verify) {
   this._verify = verify;
   this._passReqToCallback = !!options.passReqToCallback;
 
-  this._stateHandler = new SessionContentHandler('state', options.stateMaxAmount, options.stateMaxAge);
-  this._nonceHandler = new SessionContentHandler('nonce', options.nonceMaxAmount, options.nonceMaxAge);
+  this._sessionContentHandler = new SessionContentHandler(options.sessionContentMaxAmount || 10, options.sessionContentMaxAge || 3600);
```
polita: We should set the default in the config options, not here.

lovemaths: The 'options' is read from the config file. If user doesn't provide values in config, we set the default value here. You mean user must provide the value in config, if they don't provide a value, we throw an error?

lib/oidcstrategy.js
lovemaths: We use maxAmount because here we are storing things in session in server, we need to set a limit so the array won't get too large. Wilson doesn't need maxAmount because it stores things in the cookie. 

In Wilson, state string is part of the nonce string, so it makes sense to say NonceLifetime. Here the lifetime is for state, nonce, policy (for B2C), and they are 3 separated strings.

polita: While I understand that the name "sessionContentMaxAge" of the config setting might be a little more technically correct, we should be driving toward consistency of object model between our middlewares. nonceLifetime is a good name that is also correct for what it controls.

I don't understand how nonceMaxAmount (or whatever it's called) behaves. What happens if you hit the max amount but nothing is expired? Do we start kicking things out of session even though it's not expired?

lovemaths: OK I will change the name to 'nonceMaxAmount'  and 'nonceLifeTime'.

Yes, if we hit the max amount, we just kick out the oldest one even though it is not expired. In reality we have more than one nonce/state when a user opens the microsoft login page (where they can put password) on multiple tabs without putting in their password. The default number 10 should be a reasonable amount.

polita: Nit: nonceLifetime. Don't capitalize the t.

lovemaths: Changed the name to 'nonceLifetime'; removed 'maxAmount' from the code, hard coded '10' for that value in the code.

lovemaths:
@polita @brentschmaltz Is this PR ready to go? Anything else I need to change?

lib/oidcstrategy.js
```diff
@@ -190,6 +193,12 @@ function Strategy(options, verify) {
   this._verify = verify;
   this._passReqToCallback = !!options.passReqToCallback;
 
+  // For each microsoft login page tab, we generate a {state, nonce, policy, timeStamp} tuple,
+  // normally user won't keep opening microsoft login page in new tabs without putting their 
+  // password for more than 10 tabs, so we only keep the most recent 10 tuples in session.
+  // The lifetime of each tuple is 60 minutes or user specified.
+  this._sessionContentHandler = new SessionContentHandler(10, options.nonceLifetime || 3600);
```
polita: Please define the 10 as a constant with a comment on what it's for. This looks like a magic number. It's probably worth defining the 3600 as a constant too.

polita:
:shipit:

Issue #218, opened 2016-09-24
Missing email claim for B2C


tejak1:
There is no way to mention the name of a custom sign-in policy name in the '/examples/login-oidc-b2c/client_config_b2c.js' file, unlike the examples/server-oauth2-b2c/server_config.js file (policyName). I am unable to get proper claims post sign-in (especially email) for my Node app with the current configuration scope.

lovemaths:
Passport has a bug returning the email claim. We are working on the fix and we will release as soon as we can with a fix to B2C. Thanks.

tejak1:
Thank you.

lovemaths:
@tejak1 By the way, OIDC figures out the policy by itself, we don't provide policy names in the config file. User just needs to make sure they add the policy name in the route they want to authenticate, please have a look at the routes in 'views/index.ejs'. I changed the title of this issue to 'missing email claim for B2C'.

tejak1:
Yes, I have the policy name updated in the routes. I didn't know if the 'no email' found' error was thrown due to not mentioning the policy name in the configuration file also. Thanks for the clarification.

Issue #194, opened 2016-09-08
Error message for 'sub' mismatch is incorrect after redeeming 'code'


brentschmaltz:
When the 'sub' doesn't match the error message reports the 'iss' claim is mismatched.

if (subFromPrevIdToken && subFromPrevIdToken !== jwtClaims.sub)
            return self.fail('After redeeming the code, iss in id_token from authorize_endpoint does not match iss in id_token from token_endpoint');

Issue #165, opened 2016-08-10
validationConfiguration.callbackUrl should be named redirectUrl


brentschmaltz:
the name callbackUrl is not in the spec and conflicts with other libraries (wilson, asp.net).

see: http://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthRequest

polita:
This will be a breaking change.

PR #220, opened 2016-09-26
Fix OIDC B2C problems

lovemaths:
(1) Rewrote the metadata loading and the configuration code. Restructured OIDCStrategy.
(2) #188 B2C mocha tests (partially done, waiting for the AAD fix of missing nonce to add test for hybrid/code flow)
(3) #165 rename 'callbackURL' and 'returnURL' to 'redirectUrl'.
(4) #189 Extensibility to allow issuer validation when going against commend end point
(5) #194 error message for 'sub' mismatch is incorrect after redeeming 'code'
(6) #218 missing email claim for B2C

### DIFF
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -71,22 +71,17 @@ const NONCE_LIFE_TIME = 3600; // second
 const B2C_PREFIX = 'b2c_1_';
 
 function makeProfileObject(src, raw) {
-  var emails = [];
-  if (src.upn)
-    emails = [src.upn];
-  else if (src.emails)
-    emails = src.emails;
-
   return {
     sub: src.sub,
     oid: src.oid,
+    upn: src.upn,
     displayName: src.name,
     name: {
       familyName: src.family_name,
       givenName: src.given_name,
       middleName: src.middle_name,
     },
-    emails: emails,
+    emails: src.emails,
     _raw: raw,
     _json: src,
   };
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -181,9 +181,15 @@ function onProfileLoaded(strategy, args) {
  *                          How you get the authorization code and tokens back
  *
  *   - `redirectUrl`        (1) Required
- *                          (2) must be a http or https url string
+ *                          (2) must be a https url string, unless you set `allowHttpForRedirectUrl` to true
  *                          (3) Description:
  *                          The reply URL registered in AAD for your app
+ *
+ *   - `allowHttpForRedirectUrl`
+ *                          (1) Required to set to true if you want to use http url for redirectUrl
+ *                          (2) Description:
+ *                          The default value is false. It's OK to use http like 'http://localhost:3000' in the
+ *                          dev environment, but in production environment https should always be used. 
  *
  *   - `clientSecret`       (1) Required if `responseType` is 'code', 'id_token code' or 'code id_token'
  *                          (2) must be a string
@@ -240,6 +246,7 @@ function onProfileLoaded(strategy, args) {
  *     responseType: config.creds.responseType,
  *     responseMode: config.creds.responseMode
  *     redirectUrl: config.creds.redirectUrl, 
+ *     allowHttpForRedirectUrl: config.creds.allowHttpForRedirectUrl,
  *     clientSecret: config.creds.clientSecret,
  *     isB2C: config.creds.isB2C,
  *     validateIssuer: config.creds.validateIssuer,
@@ -426,11 +433,17 @@ function Strategy(options, verify) {
 
   var validatorConfiguration = {
     clientID: Validator.isNonEmpty,
-    redirectUrl: Validator.isURL,
     responseType: Validator.isTypeLegal,
     responseMode: Validator.isModeLegal,
     identityMetadata: Validator.isHttpsURL
   };
+  
+  // redirectUrl is https by default
+  if (options.allowHttpForRedirectUrl === true)
+    validatorConfiguration.redirectUrl = Validator.isURL;
+  else
+    validatorConfiguration.redirectUrl = Validator.isHttpsURL;
+
   // validator will throw exception if a required option is missing
   var validator = new Validator(validatorConfiguration);
   validator.validate(itemsToValidate);
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -31,10 +31,10 @@ const base64url = require('base64url');
 const cacheManager = require('cache-manager');
 const _ = require('lodash');
 const jws = require('jws');
-const objectTransform = require('oniyi-object-transform');
 const passport = require('passport');
 const querystring = require('querystring');
 const url = require('url');
+const urlValidator = require('valid-url');
 const util = require('util');
 
 // packages from this library
@@ -48,7 +48,6 @@ const Log = require('./logging').getLogger;
 const Metadata = require('./metadata').Metadata;
 const OAuth2 = require('oauth').OAuth2;
 const SessionContentHandler = require('./sessionContentHandler').SessionContentHandler;
-const UrlValidator = require('valid-url');
 const Validator = require('./validator').Validator;
 
 // global variable definitions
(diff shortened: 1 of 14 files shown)


lib/oidcstrategy.js
```diff
@@ -625,78 +751,71 @@ Strategy.prototype.setOptions = function setOptions(options, metadataUrl, cachek
...
 
     // jwt checks the 'nbf', 'exp', 'aud', 'iss' claims
     // there are a few other things we will check below
 
-    // check the nonce in claims
-    if (!jwtClaims.nonce || jwtClaims.nonce !== nonce)
-      return self.fail('invalid nonce');
+    // For B2C, check the policy
+    if (self._options.isB2C) {
+      if (!jwtClaims.acr || jwtClaims.acr.length <= 6 || jwtClaims.acr.substring(0,6).toUpperCase() !== 'B2C_1_')
```
polita: Please define the B2C_1_ string as a constant and then use value.length.

lib/oidcstrategy.js
```diff
@@ -786,45 +900,50 @@ Strategy.prototype._hybridFlowHandler = function hybridFlowHandler(id_token, cod
...
  * handle the response where we only get 'code' in the response
  *
- * @params {Object} code
- * @params {Object} tuple  -- state/nonce/policy/timeStamp tuple
+ * @params {Object} params
+ * @params {Object} oauthConfig
+ * @params {Object} optionsToValidate 
  * @params {Object} req
- * @params {Function} next
+ * @params {Function} nextt
```
polita: typo? Are we actually using this variable?

lib/oidcstrategy.js
```diff
@@ -836,60 +955,47 @@ Strategy.prototype._authCodeFlowHandler = function authCodeFlowHandler(code, tup
...
-    if (params.token_type !== 'Bearer') {
-      log.info('token_type received is: ', params.token_type);
-      return self.fail(`token_type received is not 'Bearer'`);
+    if (items.token_type !== 'Bearer') {
+      log.info('token_type received is: ', items.token_type);
+      return self.failWithLog(`In _authCodeFlowHandler: token_type received is not 'Bearer'`);
     }
 
-    log.info('received id_token: ', id_token);
+    log.info('received id_token: ', items.id_token);
```
polita: Are we logging the contents of the id_token? I don't think we want to do that.

lib/oidcstrategy.js
```diff
@@ -986,78 +1085,55 @@ Strategy.prototype._flowInitializationHandler = function flowInitializationHandl
...
-  let scope = config.scope;
-  if (Array.isArray(scope)) {
-    scope = scope.join(config.scopeSeparator);
+  var policy = null;
+  if (self._options.isB2C) {
+    if (!req.query.p || req.query.p === '')
+      return self.failWithLog('In _flowInitializationHandler: missing policy in the request for B2C');
+    // policy is not case sensitive. AAD turns policy to lower case.
+    policy = req.query.p.toLowerCase();
+    if (!policy.startsWith('b2c_1_'))
```
polita: Use the same constant as above.

lib/oidcstrategy.js
```diff
@@ -69,18 +68,22 @@ const NONCE_MAX_AMOUNT = 10;
 const NONCE_LIFE_TIME = 3600; // second
 
 function makeProfileObject(src, raw) {
+  var emails = [];
+  if (src.upn)
+    emails = [src.upn];
```
lovemaths: upn is always the email

lib/oidcstrategy.js
```diff
@@ -69,18 +68,22 @@ const NONCE_MAX_AMOUNT = 10;
...
+  else if (src.emails)
+    emails = src.emails;
```
lovemaths: no. For B2C we get `emails` but not `upn`; for non B2C we get `upn` but no `emails`

lib/oidcstrategy.js
lovemaths: we used to, but not anymore, will remove the 'next'

lib/oidcstrategy.js
lovemaths: why not? logging is only seen by the developers, with the id_token it is much easier for them to debug issues

lib/oidcstrategy.js
lovemaths: ok, will do it.

lib/oidcstrategy.js
```diff
@@ -144,42 +151,114 @@ function onProfileLoaded(strategy, args) {
...
- *   - `responseMode`       For login only flows we should have token passed back to us in a POST
- *   - `validateIssuer`     if you have validation on, you cannot have users from multiple tenants sign in
- *   - `passReqToCallback`  if you want the Req to go back to the calling function for other processing use this.
- *   - `nonceLifetime`      the lifetime of nonce in session, default is NONCE_LIFE_TIME
+ *   - `identityMetadata`   (1) Required
+ *                          (2) must be a https url string
+ *                          (3) Description:
+ *                          the metadata endpoint provided by the Microsoft Identity Portal that provides 
+ *                          the keys and other important info at runtime. Examples:
+ *                          <1> v1 non-common endpoint
```
brentschmaltz: should we say: tenant endpoint

lib/oidcstrategy.js
```diff
@@ -144,42 +151,114 @@ function onProfileLoaded(strategy, args) {
...
+ *                          (2) must be 'code', 'code id_token', 'id_token code' or 'id_token'
+ *                          (3) Description:
+ *                          For login only flows use 'id_token'. For accessing resources use `code id_token`, 'id_token code' or `code`
+ *
+ *   - `responseMode`       (1) Required
+ *                          (2) must be 'query' or 'form_post'
+ *                          (3) Description:
+ *                          How you get the authorization code and tokens back
+ *
+ *   - `redirectUrl`          (1) Required
```
brentschmaltz: indented too far

lib/oidcstrategy.js
lovemaths: I think it is already very clear here with the examples.

lib/oidcstrategy.js
lovemaths: modified

lib/oidcstrategy.js

lib/oidcstrategy.js

lib/oidcstrategy.js

lib/oidcstrategy.js
tushargupta51: I think for non B2c, we get `email` claim in the id token.

lib/oidcstrategy.js
tushargupta51: upn is not necessarily the email of the user. It might have `@` in it. I think this property should be called `username` instead of `emails`.

lovemaths: For AAD, upn is always in the email format like 'someone@sometenant.onmicrosoft.com'

lib/oidcstrategy.js
lovemaths: Confirmed with Danny, AAD won't return `email` claim in id_token. To get `email` claim, one needs to call graph api.

lib/oidcstrategy.js
polita: I spoke to Brent, and we assume for Wilson that the log can contain PII. This seems fine.

lib/oidcstrategy.js
polita: Agree with Brent. We should call this "v1 tenant-specific (not common) endpoint".

lib/oidcstrategy.js
tushargupta51: You are right. It is always in email format. But I am not sure if we can use that as email of the user. I saw you updated the code to set emails: src.emails. What is src here? json for id token? Is emails a new property?

lib/oidcstrategy.js
polita: @lovemaths Is this still an open issue?

lovemaths: changed it in the latest commit

polita: Thanks.

polita:
:shipit:

tushargupta51:
:shipit:  (I only reviewed makeProfileObject method)

PR #221, opened 2016-09-28
Fix xpath expression in validateSignature

xavierdutreilh:
We spot a bug in `passport-azure-ad` 2 (it affects all versions from 2.0.0 to 2.0.3).

In 0ac89df8679e9e0a2744ccc0de81a620f914a1b8, the `validateSignature` function has been rewritten and a typo on the namespace URI was made: `http://www.w3.org/2000/09/xmldsig#` became `http: //www.w3.org/2000/09/xmldsig#` (notice the extra whitespace added after `http:`).

This change broke the `validateSignature` function. The line `sig.loadSignature(signature.toString());` would always yield a `Cannot read property 'toString' of undefined` message.


lovemaths:
@xavierdutreilh Thank you for your contribution, we just merged your PR.

Issue #195, opened 2016-09-08
Remove default query support for access_tokens in bearerstrategy


brentschmaltz:
https://tools.ietf.org/html/rfc6750#section-5.3 calls out not to use query unless absolutely necessary. Query parameters are commonly logged and maintained in browser history.

PR #222, opened 2016-10-01
Some fixes to BearerStrategy

lovemaths:
(1) Issue #195 disallow bearer token in query
(2) on common endpoint, user can set validateIssuer true if they provide issuer
(3) breaking changes:
        1. add isB2C option for using B2C
	2. no longer support user provided certificate, we only allow the certificate created from AAD metadata
	3. removed tenantName option for B2C. Instead of using common endpoint then replacing the 'common' with tenantName, we now ask user to directly provide the identityMetadata with tenantName.


lib/bearerstrategy.js
```diff
@@ -129,103 +169,133 @@ function Strategy(options, verifyFn) {
...
+    options.isB2C = false;
 
-  // the default value of validateIssuer is true 
-  if (options.validateIssuer === undefined || options.validateIssuer === null) 
-    options.validateIssuer = true;
+  // turn issuer into an array
+  if (options.issuer === '')
+    options.issuer = null; 
+  if (options.issuer && !Array.isArray(options.issuer))
+    options.issuer = [options.issuer];
```
polita: What happens if you assign it an empty array?

lib/bearerstrategy.js
```diff
@@ -264,101 +336,56 @@ Strategy.prototype.authenticate = function authenticateStrategy(req) {
...
-
-        // We are replacing the common endpoint with the concrete metadata of a B2C tenant.
-        metadataURL = metadataURL
-            .replace('common', self._options.tenantName)
+      if (self._options.isB2C) {
+        params.cacheKey = 'policy: ' + self._options.policyName;
+        params.metadataURL = metadataURL
             .concat(`?p=${self._options.policyName}`)
             .concat(`&${aadutils.getLibraryProductParameterName()}=${aadutils.getLibraryProduct()}`)
             .concat(`&${aadutils.getLibraryVersionParameterName()}=${aadutils.getLibraryVersion()}`);
```
polita: Can we pull out the telemetry so we only have one piece of code that adds it? You should be able to add the ? or & as part of the line above for B2C and have a single call to .concat for non-B2C with just the "?".

lib/jsonWebToken.js
```diff
@@ -165,7 +165,7 @@ exports.verify = function(jwtString, PEMKey, options, callback) {
   if (options.allowMultiAudiencesInToken === false && Array.isArray(payload.aud) && payload.aud.length > 1)
     return done(new Error('mulitple audience in token is not allowed'));
   var payload_audience = Array.isArray(payload.aud) ? payload.aud : [payload.aud];
-  if (payload_audience.indexOf(options.audience) == -1)
+  if (payload_audience.indexOf(options.audience) == -1 && payload_audience.indexOf('spn:'+options.audience) == -1)
```
polita: What is "spn:[audience]"? Nit: need a space after the +.

lib/bearerstrategy.js
lovemaths: app will fail because of wrong issuer

lib/jsonWebToken.js
lovemaths: sometimes the `aud` is payload is 'spn:clientid', sometimes it is 'clientid'. Here it can cover both cases.

lib/bearerstrategy.js
lovemaths: Sure.

lib/bearerstrategy.js
lovemaths: now change it to null if it is empty array

lib/bearerstrategy.js
lovemaths: changed it.

lib/bearerstrategy.js
```diff
@@ -264,101 +338,53 @@ Strategy.prototype.authenticate = function authenticateStrategy(req) {
...
-
-        // use the policy name as the metadata cache key
-        cacheKey = 'policy: ' + policyName;
+      if (self._options.isB2C) {
+        params.cacheKey = 'policy: ' + self._options.policyName;
+        params.metadataURL = params.metadataURL.concat(`&p=${self._options.policyName}`)
       } else {
-        metadataURL = metadataURL.concat(`?${aadutils.getLibraryProductParameterName()}=${aadutils.getLibraryProduct()}`)
-        .concat(`&${aadutils.getLibraryVersionParameterName()}=${aadutils.getLibraryVersion()}`);
+        params.cacheKey = 'odinary';
```
polita: Is this supposed to say "ordinary"? Why is that the cache key?

lovemaths: It can be anything we like. We used the same word 'ordinary' for non-B2C in OIDCStrategy too.

polita: Did we spell it correctly there? ;)

lovemaths: I see. It was correct in OIDC. I fixed this typo in the latest commit, and updated the "v1 tenant-specific" wording.

polita:
:shipit:

PR #227, opened 2016-10-10
wrap up for 3.0.0

lovemaths:
(1) update changelog

(2) apply changes from master since 2.0.1

(3) remove the leftover redundant packages and code


CHANGELOG.md
```diff
@@ -4,26 +4,88 @@
...
 'code id_token', 'id_token code'). For all other scenarios, we do an 'id_token' fallback, since we 
 cannot obtain 'userinfo' (AAD v2 doesn't have an userinfo endpoint, and 'id_token' flow doesn't yield
  an 'access_token' to exchange for an 'userinfo' token).
 
-* changed the `returnURL` option name to `redirectUrl`.
+* `returnURL` option: this option is renamed to `redirectUrl`. `redirectUrl` can only be a https url now unless you set the
+`allowHttpForRedirectUrl` option to true.
 
-* added `isB2C` option. In order to use B2C feature, user must have this option and set it to true. 
+* `isB2C` option: this is a new option. In order to use B2C feature, user must have this option and set it to true.
```
polita: "If you are using a B2C tenant, set this option to true."

CHANGELOG.md
```diff
@@ -4,26 +4,88 @@
...
+
+* We no longer support the `certificate` option. Now we always fetch the keys from the metadata url and generate the pem key.
+
+#### B2C only
+
+* `identityMetadata`: common endpoint is no longer allowed for B2C. Tenant-specific endpoint should be used, for instance:
+`https://login.microsoftonline.com/your_B2C_tenant_name.onmicrosoft.com/v2.0/.well-known/openid-configuration` or 
+`https://login.microsoftonline.com/your_B2C_tenant_guid/v2.0/.well-known/openid-configuration`.
+
+* `isB2C`: this is a new option. In order to use B2C feature, this option is required and it must be set to true.
```
polita: Use the same wording as above.

CHANGELOG.md
lovemaths: done

CHANGELOG.md
lovemaths: done

lib/oidcstrategy.js
```diff
@@ -504,10 +498,13 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
...
   const self = this;
 
+  var resource = options && options.resourceURL;
+  var customState = options && options.customState;
+
   // 'params': items we get from the request or metadata, such as id_token, code, policy, metadata, cacheKey, etc
   var params = {}; 
   // 'oauthConfig': items needed for oauth flow (like redirection, code redemption), such as token_endpoint, userinfo_endpoint, etc
-  var oauthConfig = {};
+  var oauthConfig = { 'resource': resource, 'customState': customState};
```
Huan9: So this make the `resource` and `state` configurable for each request, right? Meaning, for each incoming request, we can pass in different value, instead one-time only config during setup time.

lib/metadata.js
```diff
@@ -37,9 +36,8 @@ function Metadata(url, authtype, options) {
   if (!url) {
     throw new Error('Metadata: url is a required argument');
   }
-  if (!authtype) {
-    throw new Error(`OIDCBearerStrategy requires an authentication type specified to metadata parser.
-      Valid types are "saml", "wsfed", or "odic"`);
+  if (!authtype || authtype !== 'oidc') {
```
SomkaPe: Does it mean that we do not want to support wsfedstrategy and saml  ? Looks like Metadata is used by them -  they call it with "saml" and "wsfed"  authtypes, do we want to remove wsfedstrategy and saml  completely ?

polita:
:shipit:

lib/metadata.js
lovemaths: yes. "saml" and "wsfed" not longer supported in 3.0.0

lib/oidcstrategy.js
```diff
@@ -992,7 +999,8 @@ Strategy.prototype._authCodeFlowHandler = function authCodeFlowHandler(params, o
     }
 
     log.info('received id_token: ', items.id_token);
-    log.info('received access_token: ', items.access_token);
+    log.info('received access_token: ', access_token);
+    log.info('received refresh_token: ', refresh_token);
```
SomkaPe: should we use debug level for sensitive info ?

lovemaths: @SomkaPe This is server side, only the developer can see it.

SomkaPe:
:shipit:

PR #234, opened 2016-10-15
Search for e-mail & break-up displayName if individual names not set

### LINKED ISSUE #174: Profile name attributes are undefined
This is probably not the right place for this since it looks like the server side thing.
When getting the profile back i get the following:
```
  displayName: 'Daniel Granatshtein',
  name: 
   { familyName: undefined,
     givenName: undefined,
     middleName: undefined },
```
The raw response contains only ```"name":"Daniel Granatshtein"```

Looking at the function makeProfileObject in oidcstrategy.js:
```
displayName: src.name,
    name: {
      familyName: src.family_name,
      givenName: src.given_name,
      middleName: src.middle_name,
    },
```
This is confusing, and requires me to break it down myself (unless i missed anything).

IMHO Ideally the names should have been send separately from the server, but as long as those are not send, the profile should not contain the name block, or it should break the raw.name in to the profile.name block.

What do you think? Have I missed anything?

### LINKED ISSUE #232: Profile Data Not Loading
I've been attempting to get the azuread-oidc strategy working today and have run into a number of issues even getting it running. It seems that a number of things have changed and documentation has not been updated.

In any case, the last (and hopefully final) issue I've run into is that the profile does not appear to populate with the email, givenName, lastName, etc. (See the screenshot):
profile_not_loaded

This is the case regardless of if I set the scope to "openid", "profile", or "email". In case it matters, my latest code is below, but I believe the "profile" should really automatically be populated by passport:
```
passport.use(new OIDCStrategy({
      callbackURL: ((config.secure && config.secure.ssl === true) ? "https" : "http") + "://" + config.domain + ":" + config.port + config['azuread-openidconnect'].callbackURL,
      clientID: config['azuread-openidconnect'].clientID,
      clientSecret: config['azuread-openidconnect'].clientSecret,
      identityMetadata: 'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration',
      responseType: 'id_token',
      responseMode: 'form_post',
      validateIssuer: false,
      redirectUrl: ((config.secure && config.secure.ssl === true) ? "https" : "http") + "://" + config.domain + ":" + config.port + config['azuread-openidconnect'].callbackURL,
      allowHttpForRedirectUrl: (config.secure && config.secure.ssl === true) ? false : true,
      scope: ['openid']
      },
      function(iss, sub, profile, accessToken, refreshToken, done) {
        console.log(JSON.stringify(profile));
        console.log('Example: Email address we received was: ', profile.email);
        var lastName = profile.displayName.split(/ +/).pop().trim();
        var firstName = profile.displayName.substr(0, profile.displayName.lastIndexof(lastName));
        // Create the user OAuth profile
        var providerUserProfile = {
          firstName: firstName,
          lastName: lastName,
          displayName: profile.displayName,
          email: profile.emails ? profile.emails[0].value : undefined,
          username: profile.username || generateUsername(profile),
          profileImageURL: (profile.id) ? '//graph.facebook.com/' + profile.id + '/picture?type=large' : undefined,
          provider: 'facebook',
          providerIdentifierField: 'id',
          providerData: providerData
        };

        // Save the user OAuth profile
        users.saveOAuthUserProfile(req, providerUserProfile, done);


      }
  ));

```
It is also worth noting that I originally tried to setup an application within Portal.Azure.com -> Azure AD -> App Registrations. I also tried setting up an application in the equivalent under the Classic portal. Both of these gave me the error, "is not supported for this API version". It was only when I created the application at apps.dev.microsoft.com that everything appeared to work (minus loading the profile).

securityvoid:
The following pull request defaults the e-mail address and name fields based on results common in an Azure AD response. This allows for a more expected result when working with Passport utilizing the standard fields used by other Passport modules.

Fixes #232  , Fixes #174

### DIFF
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -70,18 +70,62 @@ const NONCE_LIFE_TIME = 3600; // second
 // B2C policy is like 'b2c_1_policyname'. policy is not case sensitive. 
 const B2C_PREFIX = 'b2c_1_';
 
+/**
+ * validateEmail - Takes an incoming string and returns true if its a valid e-mail or false if it is not.
+ * @param email the e-mail
+ * @returns {boolean} - True if its a valid email syntax, false if it is not.
+ */
+function validateEmail(email) {
+  var re = /^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+  return re.test(email);
+}
+
+/**
+ * splitDisplayName - Takes a displayName and splits it into First, Middle, Last.
+ * @param displayName - An entire name as one string.
+ * @returns {*} - An array of size 3, 0 = First Name, 1=Middle Name, 2=Last Name
+ */
+function splitDisplayName(displayName){
+  if(!displayName)//If displayName not set, return all undefines
+    return [undefined, undefined, undefined];
+
+  var names = displayName.trim().split(/ +/);
+
+  switch(names.length){
+    case 1://1 & 3 are same, since with 1, names[1], names[2] will be undefined
+    case 3:
+      return [names[0], names[1], names[2]];
+    case 2://If only 2 its firstname/lastname
+      return [names[0], undefined, names[1]];
+    default://Anything else, assume first name is the long one.
+      var lname = names.pop();
+      var mname = names.pop();
+      return [names.join(' '), mname, lname];
+  }
+}
+
 function makeProfileObject(src, raw) {
+  //Default the E-mail if Standard Location Not Set
+  var emails = (src.emails) ? src.emails : undefined;
+  emails = (!emails && validateEmail(src.upn)) ? [{ value: src.upn }] : undefined;
+  emails = (!emails && validateEmail(src.preferred_username)) ? [{ value : src.preferred_username }] : undefined;
+
+  //If none of name details are set, Default name values based on displayName
+  var names = (!src.family_name && !src.middle_name && !src.given_name) ?
+      splitDisplayName(src.name) : [src.family_name, src.middle_name, src.given_name];
+
   return {
     sub: src.sub,
     oid: src.oid,
     upn: src.upn,
     displayName: src.name,
+    username: src.username || src.preferred_username,
     name: {
-      familyName: src.family_name,
-      givenName: src.given_name,
-      middleName: src.middle_name,
+      familyName: names[0],
+      middleName: names[1],
+      givenName: names[2],
     },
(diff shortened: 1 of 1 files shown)


securityvoid:
@lovemaths Any thoughts on this pull request? I've got another pull I'm planning on submitting on another project, integrating Azure AD, but the syntax of that pull will ultimately be based on what is determined here.

I think this will be a good enhancement, since it will allow passport-azure-ad to give a more expected response, but based on our prior conversations I'm not sure if you agree.

lib/oidcstrategy.js
```diff
@@ -70,18 +70,62 @@ const NONCE_LIFE_TIME = 3600; // second
...
+      var mname = names.pop();
+      return [names.join(' '), mname, lname];
+  }
+}
+
 function makeProfileObject(src, raw) {
+  //Default the E-mail if Standard Location Not Set
+  var emails = (src.emails) ? src.emails : undefined;
+  emails = (!emails && validateEmail(src.upn)) ? [{ value: src.upn }] : undefined;
+  emails = (!emails && validateEmail(src.preferred_username)) ? [{ value : src.preferred_username }] : undefined;
```
lovemaths: The logic of 109-111 is not correct. Each line assigns `emails` a new value, and after running line 111 `emails` will be either [{ value: src.preferred_username }] or undefined.

This library supports three different types of tenants: v1, v2 and B2C. `emails` is a claim only used by B2C tenant.

lib/oidcstrategy.js
```diff
@@ -70,18 +70,62 @@ const NONCE_LIFE_TIME = 3600; // second
...
     upn: src.upn,
     displayName: src.name,
+    username: src.username || src.preferred_username,
     name: {
-      familyName: src.family_name,
-      givenName: src.given_name,
-      middleName: src.middle_name,
+      familyName: names[0],
+      middleName: names[1],
+      givenName: names[2],
```
lovemaths: What if AAD returns the family_name, given_name etc in the claim, will the names array always match these claims?

lovemaths:
@securityvoid Can you make it part of your verify function instead of adding it to this library? We support three different types of tenants and there are some differences on what claims they support. Due to the complexity, we only propagate the original claims received from AAD back to the user. We don't want to craft any claim since that might conflict with the claim we potentially receive from AAD.

securityvoid:
@lovemaths I'd be happy to refactor and rework if you can give examples of the different types of responses you receive based on the type. The concept of defaulting the values should be solid, even if I had a bug in what I wrote.

Most people who consume the passport module are not going to know that there are three possible formats. This is going to give all sorts of inconsistent results if this module does help handle that for them. One of the big advantages of using something like Passport, is that these details are usually abstracted from the user. Typically the maker of the passport module knows about all the possible values to be returned (or at least fixes those formats when bugs are filed). Everyone else just references the default fields.

I'd really rather not reference an object which is prefixed with an underscore, aka its not standardized and the results may change. I'd much rather expand the original code for the module to standardize things, so going forward people can trust that the values, if available, will be set in the appropriate spots.

Issue #233, opened 2016-10-15
Provide documentation with more details


lovemaths:

polita:
Two things:

- Create a table in the readme with rows for V1, V2, B2C and columns for each config option with their valid values.

- API documentation

Issue #229, opened 2016-10-12
use tenent id dynamically for each request


lovemaths:
allow user to choose the tenant id you want to use for each request

polita:
@huan9 Can you clarify your scenario?

Huan9:
So I am from Sonoma(under C&E Xamarin) team. We are trying to enable grant app-only permission against user permission.
So the first step is we need to show all the subscriptions owned by the user after user auth through AAD.

User may have multiple tenants associate with his\her account. in order to get all the subscriptions associate withe the user, we actually need to go through auth flow with each tenant user has access to and use the access_code associate with different tenant to get the subscription under that tenant.

so in order to complete the flow, we actually need to pass in different tenant id per auth request.

Let me know if you have question, I am more than happy to walk through the scenario. And I actually plan to have a design review on this Friday or next Monday, if you like, I can send you the invite if that helps. 

And again, thank @lovemaths for the great help! Big Kudos for the fast response!

Issue #123, opened 2016-06-14
Question: What is the difference between OIDCStrategy and BearerStrategy. Which one should I use?


dhodgin:
The documentation could use some beefing up to explain the differences on the core readme file to explain what you get back in each authenticate callback for the 2 strategies and which use cases they should be used for.

I managed to make a sandbox app using the BearerStrategy but i haven't gotten the OIDCStrategy working in any way yet. I wanted to try to swap out the strategy in my sandbox app or possibly use both in separate routes to show and see how each is used but then i saw there are 12 config options in the setup and got lost. There are also 6 arguments in the callback function for the OIDCStrategy and no comment explaining what they are.

Even if the config options and callback arguments are commented it would still be very helpful to know the use cases for each to decide which strategy i should actually use and the difference between them.

It seems they both use OAuth and I see that the bearerstrategy expects a JWT token in the authorize header but I don't get what the OIDC 'object' is that is passed between client and server and how that model works.

Perhaps a brief intro in the readme and some links to external resources to help the end user choose appropriately. right now the readme just says here's a sample of bearer and here's a sample of oidc and that's it. very confusing.

dhodgin:
I did find this great starting point that @brandwe already wrote a few months ago.

http://cudou.com/pages/dggfdibc-passport-azure-ad-which-strategy-to-use.html

You could expand on that and it would be very helpful to users who want to make use of this library.
(I would define IdP and MEAN and perhaps elaborate a bit on how angular is doing the openID Connect part for you and how that means the server should use BearerStrategy.)

I'm using the MEAN stack and adal-angular client side so from this post I understand this to mean adal-angular is performing the openID Connect part for me and when the id_token is returned as a fragment in the url to the client that token is the JWT token. The client cannot inspect the token because it is encrypted. It can only pass it along to the server to say "This is my token for authorization, I'm already authenticated". Authentication has already happened at that point. this is why BearerStrategy should be used server side to validate the JWT token and establish if the user has access rights to this resource?

We will have each user stored in our system and look them up based on email provided from the JWT token. (I haven't gotten scopes to work properly yet and seem to only be able to get email in the JWT for Microsoft accounts, not AD accounts, in the case of it being an AD account it seems the email is the 'upn' property).

An explanation of how all these dots connect would be extremely helpful in the readme.md to new developers not experienced in all this.

polita:
Please update the readme with super duper helpful info.

starlightyear:
[...]

PR #237, opened 2016-10-21
Couple of issues

lovemaths:
(1) Issue #229, use tenant id dynamically for each request

(2) Issue #233, provide documentation with more details

(3) Issue #123, Question: what is the difference between OIDCStrategy and BearerStrategy. Which one should I use?

(4) now pass all the errors to async.waterfall and call self.fail at one place instead of everywhere

(5) updated readme and changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,15 +7,15 @@
 
 * specify tenant per request 
 
-  Now you can specify the tenant per request, using the `tenantIdOrName` option in `passport.authenticate`. More details on the usage can be found in README.md. This brings us two additional benefits:
+  Now you can specify the tenant per request, using the `tenantIdOrName` option in `passport.authenticate`. More details on the usage can be found in README.md. `tenantIdOrName` enables two features:
 
   * B2C common endpoint support
 
-    Now you can use B2C common endpoint, if you specify the tenant for each login request (in other words, a request that doesn't contain code or id_token) using the `tenantIdOrName` option.
+    Now you can use the B2C common endpoint by specifying the tenant for each login request using the `tenantIdOrName` option. A login request is any request that doesn't contain code or id_token in the responseType.
 
   * extensive issuer validation on common endpoint
     
-    If you want to validate issuer on common endpoint, previously you have to provide `issuer`in configuration. Now you can either provide `issuer`, or specify the tenant for each login request.
+    Previously, you had to provide an `issuer` value in configuration to validat the issuer on the common endpoint. Alternatively, you can now specify the tenant for each login request.
 
 # 3.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,22 @@
-<a name="3.0.0"></a>
+<a name="3.0.1"></a>
+# 3.0.1
+
+## OIDCStrategy
+
+### New features
+
+* specify tenant per request 
+
+  Now you can specify the tenant per request, using the `tenantIdOrName` option in `passport.authenticate`. More details on the usage can be found in README.md. This brings us two additional benefits:
+
+  * B2C common endpoint support
+
+    Now you can use B2C common endpoint, if you specify the tenant for each login request (in other words, a request that doesn't contain code or id_token) using the `tenantIdOrName` option.
+
+  * extensive issuer validation on common endpoint
+    
+    If you want to validate issuer on common endpoint, previously you have to provide `issuer`in configuration. Now you can either provide `issuer`, or specify the tenant for each login request.
+
 # 3.0.0
 
 ## OIDCStrategy
--- a/README.md
+++ b/README.md
@@ -12,69 +12,41 @@ issuance and validation.
 _passport-azure-ad_ has been tested to work with both [Microsoft Azure Active Directory](https://azure.microsoft.com/en-us/services/active-directory/) 
 and with [Microsoft Active Directory Federation Services](http://en.wikipedia.org/wiki/Active_Directory_Federation_Services).
 
-## Security Vulnerability in Versions < 1.4.6 and 2.0.0
+## 1. Security Vulnerability in Versions < 1.4.6 and 2.0.0
 _passport-azure-ad_ has a known security vulnerability affecting versions <1.4.6 and 2.0.0. Please update to >=1.4.6 or >=2.0.1 immediately. For more details, see the [security notice](https://github.com/AzureAD/passport-azure-ad/blob/master/SECURITY-NOTICE.MD).
 
-## Versions
-Current version - 3.0.0  
+## 2. Versions
+Current version - 3.0.1  
 Minimum  recommended version - 1.4.6  
 You can find the changes for each version in the [change log](https://github.com/AzureAD/passport-azure-ad/blob/master/CHANGELOG.md).
 
-## Contribution History
+## 3. Contribution History
 
 [![Stories in Ready](https://badge.waffle.io/AzureAD/passport-azure-ad.png?label=ready&title=Ready)](https://waffle.io/AzureAD/passport-azure-ad)
 
 [![Throughput Graph](https://graphs.waffle.io/AzureAD/passport-azure-ad/throughput.svg)](https://waffle.io/AzureAD/passport-azure-ad/metrics)
 
-## Installation
+## 4. Installation
 
     $ npm install passport-azure-ad
 
-## Usage
+## 5. Usage
 
-### Configure strategy
+This library contains two strategies, OIDCStrategy and BearerStrategy.
 
-This sample uses the OAuth2Bearer Strategy:
+OIDCStrategy is for web application login purpose using OpenID connect protocol. It works in the following way:
+ If a user is not logged in, passport sends an authentication request to AAD (Azure Active Directory), then AAD returns a login page to let the user enter their credentials. Once the credentials is authenticated by AAD, the web application will eventually get an id_token back (directly from AAD authentication endpoint, or by redeeming a code at AAD token endpoint, depending on the flow you choose). Passport then validates the id_token and propagates the claims in id_token back to the verify callback, and let the passport framework finish the remaining authentication procedures. If the whole process is successful, passport adds the user information into `req.user` and passes it to the next middleware; otherwise, passport sends back an authorized response or redirects the user to the page you specify (such as homepage or login page).
 
-```javascript
+BearerStrategy is for protecting web resource/api purpose using Bearer Token protocol. It works in the following way:
+ User sends a request to the protected web api, and the request is supposed to contain an access_token in either authorization header or body. Passport extracts and validates the access_token, and propagates the claims in access_token to the verify callback and let the passport framework finish the remaining authentication procedure. If the whole process is successful, passport adds the user information into `req.user` and passes it to the next middleware, which is usually the business logic of the web resource/api; otherwise, passport sends back an authorized response.
 
-// We pass these options in to the ODICBearerStrategy.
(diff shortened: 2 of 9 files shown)


polita:
Please create one PR per issue in the future.

CHANGELOG.md
```diff
@@ -1,4 +1,22 @@
...
+<a name="3.0.1"></a>
+# 3.0.1
+
+## OIDCStrategy
+
+### New features
+
+* specify tenant per request 
+
+  Now you can specify the tenant per request, using the `tenantIdOrName` option in `passport.authenticate`. More details on the usage can be found in README.md. This brings us two additional benefits:
```
polita: Instead of "This brings us two additional benefits:", "tenantIdOrName enables two features:"

CHANGELOG.md
```diff
@@ -1,4 +1,22 @@
...
+
+  * B2C common endpoint support
+
+    Now you can use B2C common endpoint, if you specify the tenant for each login request (in other words, a request that doesn't contain code or id_token) using the `tenantIdOrName` option.
```
polita: Now you can use the B2C common endpoint by specifying the tenant for each login request using the 'tenantIdOrName' option. A login request is any request that doesn't contain code or id_token in the responseType

CHANGELOG.md
```diff
@@ -1,4 +1,22 @@
...
+
+  * extensive issuer validation on common endpoint
+    
+    If you want to validate issuer on common endpoint, previously you have to provide `issuer`in configuration. Now you can either provide `issuer`, or specify the tenant for each login request.
```
polita: Previously, you had to provide an 'issuer' valud in configuration to validat the issuer on the common endpoint. Alternatively, you can now specify the tenant for each login request. <Please insert a short description for how or link to a sample.>

README.md
```diff
@@ -12,69 +12,41 @@ issuance and validation.
 _passport-azure-ad_ has been tested to work with both [Microsoft Azure Active Directory](https://azure.microsoft.com/en-us/services/active-directory/) 
 and with [Microsoft Active Directory Federation Services](http://en.wikipedia.org/wiki/Active_Directory_Federation_Services).
 
-## Security Vulnerability in Versions < 1.4.6 and 2.0.0
+## 1. Security Vulnerability in Versions < 1.4.6 and 2.0.0
```
polita: Why are you numbering these headers? They're not part of a list.

README.md
```diff
@@ -12,69 +12,41 @@ issuance and validation.
...
-## Installation
+## 4. Installation
 
     $ npm install passport-azure-ad
 
-## Usage
+## 5. Usage
 
-### Configure strategy
+This library contains two strategies, OIDCStrategy and BearerStrategy.
```
polita: "...two strategies: OIDCStrategy and BearerStrategy"

README.md
```diff
@@ -12,69 +12,41 @@ issuance and validation.
...
 
-This sample uses the OAuth2Bearer Strategy:
+OIDCStrategy is for web application login purpose using OpenID connect protocol. It works in the following way:
```
polita: Please fix up the grammar in this section. You can use Word's grammar checker to help with some errors.

README.md
```diff
@@ -12,69 +12,41 @@ issuance and validation.
...
+ If a user is not logged in, passport sends an authentication request to AAD (Azure Active Directory), then AAD returns a login page to let the user enter their credentials. Once the credentials is authenticated by AAD, the web application will eventually get an id_token back (directly from AAD authentication endpoint, or by redeeming a code at AAD token endpoint, depending on the flow you choose). Passport then validates the id_token and propagates the claims in id_token back to the verify callback, and let the passport framework finish the remaining authentication procedures. If the whole process is successful, passport adds the user information into `req.user` and passes it to the next middleware; otherwise, passport sends back an authorized response or redirects the user to the page you specify (such as homepage or login page).
 
-```javascript
+BearerStrategy is for protecting web resource/api purpose using Bearer Token protocol. It works in the following way:
```
polita: grammar

lib/oidcstrategy.js
```diff
@@ -620,38 +619,62 @@ Strategy.prototype.collectInfoFromReq = function(params, req, next) {
...
-  if (self._options.isB2C) {
-    metadataUrl = metadataUrl.concat(`&p=${params.policy}`);
-    params.cachekey = 'policy: ' + params.policy;  // B2C metadata cachekey
-    log.info(`B2C metadataUrl is: ${metadataUrl}`);
-  } else {
-    params.cachekey = 'ordinary'; // non B2C metadata cachekey
-    log.info(`metadataUrl is: ${metadataUrl}`);
+  // if we are using common endpoint and we are given the tenantIdOrName, let's replace it
+  if (self._options.isCommonEndpoint && params.tenantIdOrName) {
+    metadataUrl = metadataUrl.replace('/common/', `/${params.tenantIdOrName}/`);
```
polita: Are we doing any simple validation on tenantIdOrName or ensure it's valid in a URL? For instance, ensuring there's not whitespace?

lib/oidcstrategy.js
```diff
@@ -620,38 +619,62 @@ Strategy.prototype.collectInfoFromReq = function(params, req, next) {
...
+    log.info(`we are replacing 'common' with the tenantIdOrName ${params.tenantIdOrName}`);
   }
 
+  // add policy for B2C
+  if (self._options.isB2C)
+    metadataUrl = metadataUrl.concat(`&p=${params.policy}`);
```
polita: Same question on validating the policy name makes sense as part of a URL.

test/Chai-passport_test/oidc_dynamic_tenant_test.js
```diff
@@ -0,0 +1,137 @@
...
+
+  var testPrepare = function(validateIssuer, issuer, tenantIdOrName, isB2C, policy) {
+  	return function(done) {
+      options.validateIssuer = validateIssuer;
+      options.issuer = issuer;
+      options.isB2C = isB2C;
+
+      var testStrategy = new OIDCStrategy(options, function(profile, done) {});
+
+  		chai.passport
```
polita: Indentation looks wrong here.

CHANGELOG.md
lovemaths: done

CHANGELOG.md
lovemaths: done

CHANGELOG.md
lovemaths: done

README.md
lovemaths: I numbered them so the structure of the readme is more clear

README.md
lovemaths: done, with the help from Rohit on grammars

lib/oidcstrategy.js
lovemaths: If the URL is invalid, authentication will fail in the metadata fetching stage

lib/oidcstrategy.js
lovemaths: same as above

lib/oidcstrategy.js
polita: I understand that's where it will fail if we don't validate. I'm suggesting we should validate before then so we can give a better error message if it's set to something invalid. This is true for both tenantIdOrName and the policy name.

README.md
```diff
@@ -12,69 +12,42 @@ issuance and validation.
...
 
-```javascript
+BearerStrategy uses Bearer Token protocol to protect web resource/api. It works in the following manner:
+User sends a request to the protected web api which contains an access_token in either the authorization header or body. Passport extracts and validates the access_token, and propagates the claims in access_token to the verify callback and let the framework finish the remaining authentication procedure. On successful authentication, passport adds the user information to `req.user` and passes it to the next middleware, which is usually the business logic of the web resource/api. In case of error, passport sends back an unauthorized response.
```
SomkaPe: Should we mention  sessions here explicitly ? 
In both cases session support is optional, 
but provided description implies that session in supported for OIDC  and not supported for bearer strategy.

For OIDCStrategy  we might add that 
common scenario is accessing a web application via a browser, 
So, after successful authentication, Passport will establish a persistent login session.


Typical use case for Berarer startegy is protection of some web API where   credentials to be supplied with each request, so parameter (session : false) should be used for passport.authenticate().
What if it is set to true ? I guess token  will not be validated if there is a session. Is  (session: true) real use case for bearer strategy ?

lib/oidcstrategy.js
```diff
@@ -359,36 +359,34 @@ function Strategy(options, verify) {
   if (options.isB2C !== true)
     options.isB2C = false;
 
-  // common endpoint is not allowed for B2C
-  if (options.isCommonEndpoint && options.isB2C) {
-    throw new Error(`Cannot use common endpoint for B2C. Please use your tenant name or tenant guid.`);
-  }
+  // common endpoint will not work for B2C unless they provide the tenant name or tenant guid in 'passport.authenticate'
+  if (options.isCommonEndpoint && options.isB2C)
+    log.info(`In order to use common endpoint for B2C, please provide your tenant name or tenant guid when you call 'passport.authenticate'.`);
```
SomkaPe: should we add check for tenant name/giud  and than throw error ?
 is it kind of advise ? it will be written to log for the case when tenant name/giud  provided and not provided, so it does not clearly reflect execution  of ther program

README.md
lovemaths: where did I say session is supported for OIDC but not supported for bearer?

passport.authenticate doesn't care if there is already a session or not, it validates token and adds/overrides `req.user` field. All { session: true or false } does is to tell passport if req.user should be saved in session. You can look at the `ensureAuthenticated` function in the sample to have some idea how passport works.

lib/oidcstrategy.js
lovemaths: @polita Now we have it.

lib/oidcstrategy.js
SomkaPe: so, what is the purpose of that 362-364 ?

lovemaths: @SomkaPe Show user this information.

README.md
SomkaPe: where did I say session is supported for OIDC but not supported for bearer -> It implies from the description :  
OIDCStrategy
"If a user is not logged in" implies sessions are supported, But there is no info in description that Passport will establish a persistent login session (only  about adding  the user information to `req.user`)

lovemaths: Why "If a user is not logged in" implies sessions are supported? If a user comes to the web app for the first time, or the user logged out, then the user is not logged in. In term of session, this means after the request goes through session middleware and comes to passport, passport doesn't see `req.user` in session.

Passport.authenticate only does the authentication work and adds user info to `req.user`, the rest (like save user to session) is done by passport framework and the web app.

lovemaths: For bearer, if there is no token, passport.authenticate fails the authentication regardless of the session. passport.authenticate is not used to check if a user is in session or not. If user wants to use the user in session, they should check `req.user` before they call passport.authenticate. Look at how `ensureAuthenticate` function in the sample to see how it works.

SomkaPe: session is supported - I mean strategy is configured with session : true.
My point was to mention in description sessions to emphasize that oidcStrategy is typically configured with session : true , bearer strategy - session false

lovemaths: Whether they want to use session true or false is totally up to the user. I will just mention the default value for session is true.

lib/oidcstrategy.js
lovemaths: Fine. Deleted the log.

polita:
:shipit: when Petro agrees.

SomkaPe:
:shipit:

Issue #239, opened 2016-10-23
Problems with signin  in the updated sample


SteenMolberg:
I have tested out the updated sample ([B2C-WebApp-OpenIDConnect-NodeJS](https://github.com/AzureADQuickStarts/B2C-WebApp-OpenIDConnect-NodeJS)) now, but have a lot of problems with the login. I keep getting this error:
"authentication failed due to: In _validateResponse: jwt is not active"

Can you help me out resolving what is going up?

Following the log it seems like the flow is working OK. I can enter the username and password and redirected back to the app. Following the log and the decoded token the corrected user seem to be returned, but still I get the error above, 

The signup seems to work OK, but when return to the application I get the same error. Still in the decoded token log I can see the new user is returned.

I'm using the default session store for my testing of the sample.

Hope you can help me out. Thanks.

lovemaths:
"jwt is not active" means the id_token received from AAD is not valid util some time in the future. Seems either we get a bad id_token from AAD or there is a difference between the two clocks, I will see if I can reproduce and let you know. Thanks.

SteenMolberg:
Thks @lovemaths 
let me know if you need any info from me.

Hmm... you say "not valid util some time in the future"... could this be a timezone issue? I'm at UTC+1

lovemaths:
@SteenMolberg I am not sure what the issue is. I cannot reproduce the issue, the sample works fine for me. In lib/jsonWebToken.js, can you replace line 192:
```
      return done(new Error('jwt is not active'));
```
with
```
      return done(new Error(`jwt is not active, nbf: ${payload.nbf}, now: ${Date.now()}`));
```
This can tell me how much they differ when the error occurs next time (note ' is changed to ` in the error  message). Also, please let me know what responseType you are using. Thanks.

SteenMolberg:
@lovemaths 
So after I made the small codeschange you requested I tested this again last night. To my surprise everything ran as expected. I logged in, cretaed a new user, logged out, changed profile. Several times. No errors. Strange.
This I repeated my testing this morning and logged in and out with diffenrent users. Everything seemed fine. But then I went to Reset password and canceled that again. Then I wanted to login again but it failed.

[...]

PR #240, opened 2016-10-27
Issue #239 Problems with signin in the updated sample

lovemaths:
AAD rounds the current time up to the nearest integer for the nbf value, but we round the current time down when we validate the nbf value. This fails the authentication because we thought the current time is before nbf and the token is not active yet.


polita:
:shipit:

PR #241, opened 2016-11-03
update changelog and add fix to nbf validation

lovemaths:
(1) updated changelog with the list of fixed bugs
(2) allow 1 more second for the nbf validation


SomkaPe:
:shipit:

PR #244, opened 2016-11-07
fix for 3.0.2

lovemaths:
(1) updated version to 3.0.2 in package.json
(2) removed oniyi-object-transform dependency
(3) set clock skew to 5 minutes as Wilson does. Will make this configurable in the future.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -26,8 +26,11 @@
 // constants that are not strategy specific
 
 var CONSTANTS = {};
+
 CONSTANTS.POLICY_REGEX = /^b2c_1_[0-9a-z._-]+$/i;    // policy is case insensitive
 CONSTANTS.TENANTNAME_REGEX = /^[0-9a-zA-Z]+.onmicrosoft.com$/;
 CONSTANTS.TENANTID_REGEX = /^[0-9a-zA-Z-]+$/;
+
+CONSTANTS.CLOCK_SKEW = 300; // 5 minutes
 
 module.exports = CONSTANTS;
--- a/lib/jsonWebToken.js
+++ b/lib/jsonWebToken.js
@@ -24,6 +24,7 @@
 'use restrict';
 
 const jws = require('jws');
+const constants = require('./constants');
 
 /* Verify the token and return the payload
  *
@@ -178,7 +179,7 @@ exports.verify = function(jwtString, PEMKey, options, callback) {
   if (typeof payload.exp !== 'number')
     return done(new Error('invalid exp value in payload'));
   if (!options.ignoreExpiration) {
-    if (Math.floor(Date.now() / 1000) >= payload.exp) {
+    if (Math.floor(Date.now() / 1000) >= payload.exp + constants.CLOCK_SKEW) {
       return done(new Error('jwt is expired'));
     }
   }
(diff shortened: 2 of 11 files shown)


lib/jsonWebToken.js
```diff
@@ -178,17 +179,17 @@ exports.verify = function(jwtString, PEMKey, options, callback) {
   if (typeof payload.exp !== 'number')
     return done(new Error('invalid exp value in payload'));
   if (!options.ignoreExpiration) {
-    if (Math.floor(Date.now() / 1000) >= payload.exp) {
+    if (Math.floor(Date.now() / 1000) >= payload.exp + constants.CLOCK_SKEW) {
```
brentschmaltz: is Date.now() in UTC?

lovemaths: The Date.now() method returns the number of milliseconds elapsed since 1 January 1970 00:00:00 UTC.

lib/jsonWebToken.js
```diff
@@ -178,17 +179,17 @@ exports.verify = function(jwtString, PEMKey, options, callback) {
...
       return done(new Error('jwt is expired'));
     }
   }
 
   // (5) nbf
   //   - check if it exists
   if (payload.nbf) {
-    if (typeof payload.nbf !== 'number')
+    if (typeof payload.nbf !== 'number' || payload.nbf >= payload.exp)
```
brentschmaltz: It would be helpful to users if the error message indicated that nbf > exp OR nbf was not a number

lib/metadata.js
```diff
@@ -71,19 +70,15 @@ Metadata.prototype.updateOidcMetadata = function updateOidcMetadata(doc, next) {
...
-    },
-    pick: [
-      'end_session_endpoint',
-      'issuer',
-      'token_endpoint',
-      'userinfo_endpoint',
-    ],
-  });
+  var oidc = {};
+  oidc.algorithms = doc.id_token_signing_alg_values_supported;
```
brentschmaltz: any particular reason for a name difference between 'doc' and 'oidc'? For example why auth_endpoint and authorization_endpoint (which is the name in the spec). I would stick with the names in the spec, otherwise we are always translating. Sometimes, we may have overlap.  Check out the names in: https://login.microsoftonline.com/common/.well-known/openid-configuration

lib/constants.js
```diff
@@ -26,8 +26,11 @@
 // constants that are not strategy specific
 
 var CONSTANTS = {};
+
 CONSTANTS.POLICY_REGEX = /^b2c_1_[0-9a-z._-]+$/i;    // policy is case insensitive
 CONSTANTS.TENANTNAME_REGEX = /^[0-9a-zA-Z]+.onmicrosoft.com$/;
 CONSTANTS.TENANTID_REGEX = /^[0-9a-zA-Z-]+$/;
 
+CONSTANTS.CLOCK_SKEW = 300; // 5 minutes
```
brentschmaltz: Please open an issue to make CLOCK_SKEW configurable.

lib/jsonWebToken.js
brentschmaltz: Do you have any tests for this?

lovemaths: The time is UTC. We have tests for expiration.

lib/metadata.js
lovemaths: I changed the auth_endpoint to authorization_endpoint everywhere in the second commit.

lib/jsonWebToken.js
lovemaths: changed in the second commit.

lib/metadata.js
lovemaths: I will keep the 'algorithms' one since it is a more natural name than 'id_token_signing_alg_values_supported'. In the library we used 'algorithms' everywhere and we didn't do any translation of that.

Issue #248, opened 2016-11-10
End_to_end_test showing up in test folder


stuartlong:
There is an "End_to_end_test" folder showing up under passport-azure-ad/test/ whenever I do an npm install for 3.0.2. I can't find a trace of it in the source code and it's causing me deployment issues on Windows because of  path too long issues (it's nesting it's node_module even though I'm on npm version 3.7.5).

On a related note, why are any of the tests included in the npm package in the first place?

lovemaths:
@stuartlong This is weird, I didn't have that folder in my master branch when I published it. I will do a new release tomorrow. Thanks for letting me know.

PR #249, opened 2016-11-10
Issue #248 End_to_end_test showing up in test folder

lovemaths:
v3.0.2 npm package has this issue. We update the version and re-publish the package.

v3.0.2 npm package contains an End_to_end_test folder, but master branch doesn't have it.
This is due to a git bug. End_to_end_test folder is in sijliu/MQ branch, if I open
passport-azure-ad in sublime, modify some files in End_to_end_test, then do a
"git reset --hard" and switch back to the master branch, then the End_to_end_test folder
will show up in the local master branch. This git behavior is incorrect, that folder
should not show up.

Even though that folder shows up, if I do a 'git status', it says the local master is the
same as the remote master branch, so git doesn't recognize it. However when I do a npm
publish, npm considers the End_to_end_test folder part of the master branch, and this
causes the diffference between the master branch and the v3.0.2 npm package.

This suggests that when we do a npm publish, we should always make a fresh clone from
github and publish there. Switching branch using git sometimes yields unexpected results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 <a name="3.0.1"></a>
+# 3.0.3
+
+## Bug fixes
+
+* [#248](https://github.com/AzureAD/passport-azure-ad/issues/248) End_to_end_test showing up in test folder
+
 # 3.0.2
 
 ## Changes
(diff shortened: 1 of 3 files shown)


brentschmaltz:
:shipit:

Issue #254, opened 2016-11-17
passReqToCallback does not work with bearer strategy


jes5e:
When 'passReqToCallback' is set to true, the callback passed in to a new strategy should get called with the three param option (req, user, done).  It does not.  Instead the two param option is always called (user, done).

jes5e:
I just submitted a pull request for this

lovemaths:
Delivered in 3.0.4

PR #255, opened 2016-11-17
in the bearerstrategy jwtVerify function, self.passReqToCallback was …

jes5e:
Fixes Issue #254 - passReqToCallback does not work with bearer strategy

When 'passReqToCallback' is set to true, the callback passed in to a new strategy should get called with the three param option (req, user, done).  It does not.  Instead the two param option is always called (user, done).

In the bearerstrategy jwtVerify function, self.passReqToCallback was being used instead of self._options.passReqToCallback.  This cause the callback passed in to a new strategy to get called with the two param option even when "passReqToCallback" is set to true.

### DIFF
--- a/lib/bearerstrategy.js
+++ b/lib/bearerstrategy.js
@@ -293,7 +293,7 @@ Strategy.prototype.jwtVerify = function jwtVerifyFunc(req, token, metadata, opti
 
     log.info('In Strategy.prototype.jwtVerify: VerifiedToken: ', verifiedToken);
     
-    if (self.passReqToCallback) {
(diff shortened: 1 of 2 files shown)


lovemaths:
@jes5e Thank you for your contribution. Your PR looks good to me. 

We normally do fixes in dev branch and merge dev into master after we get all the fixes. Can you do a PR against dev branch instead of master branch? I will merge your PR then. Thanks.

jes5e:
Sure. I'll fix the pull request later tonight. Thanks a lot!

Issue #264, opened 2016-12-05
OIDC authentication fails when oauth token_type is 'bearer' and not 'Bearer'


lindblad:
Got a problem for OIDC strategy on line 1030 in oidcstrategy.js
    // token_type must be 'Bearer'
    if (items.token_type !== 'Bearer') {
      log.info('token_type received is: ', items.token_type);
      return next(new Error(`In _authCodeFlowHandler: token_type received is not 'Bearer'`));
    }
Our AD returns 'bearer' token_type, not 'Bearer'. Is it possible to change the check to
if (items.token_type.toLowerCase() !== 'bearer') {
?

lovemaths:
@lindblad Yes we can do that in the next release. I will make a PR for it.

lovemaths:
Delivered in 3.0.4

PR #265, opened 2016-12-05
Issue #264 make token_type case insensitive for OIDCStrategy

lovemaths:
We used to only allow 'Bearer' for token_type, now we make it case insensitive and allow something like 'bearer'.


SomkaPe:
:shipit:

PR #266, opened 2016-12-05
Changes token type check to be either 'Bearer' or 'bearer'

lindblad:


lovemaths:
@lindblad I already made a PR with unit tests so I will close this PR. Thank you for your contribution!

Issue #256, opened 2016-11-18
Support 'tfp' for B2C


lovemaths:
The policy in B2C token is stored in 'acr' claim. Now this is configurable in AAD portal, user can choose either 'tfp' or 'acr' claim to store the policy. We don't have code to support 'tfp' yet.

lovemaths:
Delivered in 3.0.4

PR #269, opened 2016-12-08
Issue #256 Support 'tfp' for B2C

lovemaths:
When we create a B2C policy, the policy is used to be stored in 'acr' claim in id_token. Now this is configurable, and policy is stored in 'tfp' instead by default.


SomkaPe:
:shipit:

Issue #172, opened 2016-08-19
ResponseMode should be optional in OIDC requests 


brentschmaltz:
https://github.com/AzureAD/passport-azure-ad/blob/master/lib/validator.js#L70
This code requires a responseMode, but responseMode is optional. It is not recommended to send it when it equals the default.

http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

polita:
ResponseMode should support a 'default' or 'not set' value, which specifies that the responseMode shouldn't be sent and the app's registered responseMode is used.

pkanher617:
This will be supported in `@azure/msal-node` and the future passport library. Closing for now, tracking in our backlog.

PR #272, opened 2016-12-16
Issue #172 ResponseMode should be optional in OIDC requests

lovemaths:
The default responseMode for 'code' flow is 'query', for hybrid/implicit is 'fragment'. Since we don't support 'fragment', we only need to not send 'query' response_mode in case of 'code' flow.

### DIFF
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -425,6 +425,13 @@ function Strategy(options, verify) {
   if (options.responseType === 'id_token code')
     options.responseType = 'code id_token';
 
+  // For authorization code flow, the default response_mode is 'query'. If response_mode is
+  // not provided, we set it to 'query'.
+  // For hybrid and implicit flow, the default response_mode is 'fragment'. However, we do
+  // not support 'fragment', so user has to provide the response_mode.
+  if (!options.responseMode && options.responseType === 'code')
+    options.responseMode = 'query';
+
   var itemsToValidate = {};
   aadutils.copyObjectFields(options, itemsToValidate, ['clientID', 'redirectUrl', 'responseType', 'responseMode', 'identityMetadata']); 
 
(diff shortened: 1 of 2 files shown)


polita:
I don't quite understand. I think you're saying that, if the flow is hybrid or implicit, the app developer needs to explicitly set the response_mode to 'query'. If they don't explicitly set response_mode, the assumed response_mode, 'fragment', isn't supported, so we error. Is that right?

lovemaths:
@polita Yes that's right.

Behavior before the change:  
(1) response_mode must be set to either 'query' or 'form_post' explicitly by the app developer regardless of the flow, otherwise we throw in the strategy constructor when we validate response_mode.
(2) we always send the response_mode in the request to AAD

The changes:
(1) For authorization code flow, if app developer doesn't set response_mode, we don't throw now, we set it to 'query'.
(2) we don't send response_mode=query for authorization code flow.

polita:
What problem does this solve? 

It's confusing to me what the default behavior is. We shouldn't set a default on one property based on the value of another property. I recommend we don't change the current behavior unless there's a clear win here, and "in some cases you don't have to set the property" doesn't seem like a clear win.

Why do we want to drop the response_mode=query from the request? Is there a reason to not set it explicitly?

lovemaths:
@polita  The spec says it is not recommended to send the response mode if it is the default one.

"response_mode
OPTIONAL. Informs the Authorization Server of the mechanism to be used for returning parameters from the Authorization Endpoint. This use of this parameter is NOT RECOMMENDED when the Response Mode that would be requested is the default mode specified for the Response Type."

That's why we want to drop the response_mode=query for authorization code flow, since 'query' is the 'default' one for authorization code flow according to the spec.

lovemaths:
@polita The issue solved doesn't seem to be interesting or important, closing it...

PR #279, opened 2017-01-19
Fix Jenkins failure

lovemaths:
(1) allow more time by setting a timeout for tests in Chai-passport_test
(2) allow more time for the signin key to show up in End_to_end_test. Remove the unnecessary click of the 'continue' button in b2c updateprofile page and the 'cancel' button in b2c signup page.


SomkaPe:
:shipit:

Issue #285, opened 2017-03-13
express-session free support


lovemaths:
Now we store state, nonce, etc into express session, and retrieve them to validate these parameters in the response and id_token. What we want to achieve:

(1) provide alternative approach: encrypt these stuff and put into cookie, so express-session won't be necessary in case of { session: false }

(2) provide public api: so user can write their own functions to handle state, nonce, etc storage and retrieval.

lovemaths:
Shipped in v3.0.6.

PR #288, opened 2017-03-28
Issue #285 express-session free support

lovemaths:
We used to save state and nonce in session, so session is always required. Now we can encrypt and decrypt
state and nonce, and put them into cookie, so session is no longer mandatory.

We provided two more options: useCookieInsteadOfSession, cookieEncryptionKeys. To use cookie, first set the first option to true.
For the second option, provide an array of { key: '...', iv: '...' } items. For each item, key is any string
of size 32, and iv is any string of size 12. We will use the first item (with AES-256-GCM algorithm) to encrypt cookie, but we try every
item to decrypt cookie, so you can easily do key rollover.

### DIFF
--- a/lib/cookieContentHandler.js
+++ b/lib/cookieContentHandler.js
@@ -1,0 +1,159 @@
+/**
+ * Copyright (c) Microsoft Corporation
+ *  All Rights Reserved
+ *  MIT License
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this
+ * software and associated documentation files (the "Software"), to deal in the Software
+ * without restriction, including without limitation the rights to use, copy, modify,
+ * merge, publish, distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+ * OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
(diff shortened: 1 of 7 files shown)


test/End_to_end_test/oidc_cookie_test.js
```diff
@@ -0,0 +1,154 @@
...
+var service = chromedriver.get_service();
+var webdriver = chromedriver.webdriver;
+var By = webdriver.By;
+var until = webdriver.until;
+var base64url = require('base64url');
+
+var chai = require('chai');
+var expect = chai.expect;
+
+const TEST_TIMEOUT = 500000; // 30 seconds
```
SomkaPe: looks like 500 seconds

SomkaPe:
:shipit:

lib/cookieContentHandler.js
```diff
@@ -0,0 +1,159 @@
...
+
+  if (!cookieEncryptionKeys || !Array.isArray(cookieEncryptionKeys) || cookieEncryptionKeys.length === 0)
+    throw new Error('CookieContentHandler: cookieEncryptionKeys must be a non-emptry array');
+
+  for (var i = 0; i < cookieEncryptionKeys.length; i++) {
+    var item = cookieEncryptionKeys[i];
+    if (!item.key || !item.iv)
+      throw new Error(`CookieContentHandler: array item ${i+1} in cookieEncryptionKeys must have the form { key: , iv: }`);
+    if (item.key.length !== 32)
+      throw new Error(`CookieContentHandler: key number ${i+1} is ${item.key.length} bytes, expected: 32 bytes`);
```
cadesalaberry: @lovemaths 

Why would the key need to be exactly 32 bytes ?

The [cookie-session](https://www.npmjs.com/package/cookie-session) seems to accept secret/keys of any size.

lib/oidcstrategy.js
```diff
@@ -542,6 +573,7 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
   var prompt = options && options.prompt;
   var extraAuthReqQueryParams = options && options.extraAuthReqQueryParams;
   var extraTokenReqQueryParams = options && options.extraTokenReqQueryParams;
+  var response = options && options.response;
```
cadesalaberry: @lovemaths `options.response` is always undefined in my case.

In express 4, response can be accessed with `req.res`. Was this PR developed with this in mind ?

frantzyy:
Would it be possible to see an example of how to use this feature? I've searched everywhere and haven't found one. I understand how to setup up the config with the exception of do I need to specifically call the /token endpoint  or will passport-azure-ad do this for me. If so where will the jwt/token be available? in the strategy callback or response of the redirect?

Issue #301, opened 2017-05-18
Error: a key with kid %s cannot be found


oreporan:
According to passport documentation you can authenticate against 2 strategies like so:

`passport.authenticate([StrategyA, StrategyB], {session: false})`

However when StrategyA is `passport-azure-ad` BearerStrategy, and the token on the request corresponds to StrategyB. It never reached StrategyB because an exception is thrown:

```js
Error: a key with kid %s cannot be found
    at Metadata.generateOidcPEM (/.../node_modules/passport-azure-ad/lib/metadata.js:146:11)
    at Strategy.jwtVerifyFunc [as jwtVerify] (/.../node_modules/passport-azure-ad/lib/bearerstrategy.js:291:23)
    at async.waterfall (/.../node_modules/passport-azure-ad/lib/bearerstrategy.js:462:19)
```
There seems to be no `err` variable in the BearerStrategy:
```js
const clientAuthenticationStrategy = new BearerStrategy(options,
    (token, done) => {
        if (!token) return done("no token", null)
        return done(null, token);
    });
```

lovemaths:
@oreporan Thank you for your contribution, we will add your fix to our code base.

oreporan:
Any eta on this merge?

lovemaths:
@oreporan Sorry for the delay. Could you make your PR against our `dev` branch instead of `master`? I will merge your PR then. Thank you!

oreporan:
Yes, I could do that. 
Though for the future, you should probably make dev your default branch or else anyone who makes a pull request will always clone the wrong branch..

<img width="830" alt="screen shot 2017-06-10 at 9 56 57" src="https://user-images.githubusercontent.com/2146999/27000834-34dc9546-4dc3-11e7-9b79-4258a181f6d8.png">

lovemaths:
@oreporan Released new version with the change in npm.

PR #302, opened 2017-05-18
add trycatch

oreporan:
Linked with issue #301 

Solution: Add try/catch


lovemaths:
@oreporan Could you please make this PR against `dev` branch instead? `master` branch is for the latest release, we don't touch that branch until we do release. Thank you!

PR #315, opened 2017-06-10
switch to dev

oreporan:
For Issue #301


oreporan:
Can you notify when this change is released please?

lovemaths:
@oreporan We are planning on the next release, there might be a couple of more things we want to add before releasing it. I will let you know once we have a date. Thank you.

PR #319, opened 2017-06-16
catch exception thrown for missing kid

lovemaths:
generateOidcPEM throws an error if kid is missing, so we have to use try catch here.


rohitnarula7176:
:shipit:

PR #348, opened 2017-12-11
Load cookies from header

ple16:
Loads the cookies from the header as req.cookie is a function in Koa


ple16:
Totally wrong fork

PR #378, opened 2018-04-17
Update README.md

kaburkett:
Section 8 actually has the sample links :)


brandwe:
Already fixed by another PR.

PR #402, opened 2018-10-29
README.md updated

ShekharNain:
`passReqToCallback` config description updated in README file making that as a conditional property rather than the mandatory one.


brandwe:
Thank you @ShekharNain

PR #403, opened 2018-11-07
OIDC- fix email retrieval for passport callback

### LINKED ISSUE #384: Inconsistent profile information?
Hi

I may be doing it wrong, but I can't get the profile information to work consistently. I am trying to add a series of social login buttons to my app and I am setting it up like so:

```

version: 3.0.12

passport.use(new OIDCStrategy({
        identityMetadata: 'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration',
        clientID: ....,
        responseType: 'id_token',
        responseMode: 'form_post',
        redirectUrl: .....,
        allowHttpForRedirectUrl: true,
        clientSecret: .....,
        validateIssuer: false,
        isB2C: false,
        issuer: null,
        passReqToCallback: false,
        scope: ['profile','email'],
        useCookieInsteadOfSession: false,
    },
```

Now in the case when I log in using a personal hotmail account the e-mail address ends up in:

profile._json.email

With an organisational account:

e-mail is in profile._json.preferred_username

I can take account of this in my code, but every other passport login strategy is filling in  profile.emails[0].value like the passport docs say they should.

tzachshabtay:
Fixes #384

Note: I have no idea if the `src.emails` is actually getting filled for some people?
It hasn't from what I've seen and from the issue above, but I kept it to be safe.

### DIFF
--- a/lib/oidcstrategy.js
+++ b/lib/oidcstrategy.js
@@ -81,7 +81,7 @@ function makeProfileObject(src, raw) {
       givenName: src.given_name,
       middleName: src.middle_name,
     },
-    emails: src.emails || [{value: src.email || src.preferred_username }],,
+    emails: src.emails || [{value: src.email || src.preferred_username }],
     _raw: raw,
     _json: src,
   };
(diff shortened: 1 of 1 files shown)


jasonnutter:

Issue #432, opened 2019-07-23
Error: In Strategy.prototype.jwtVerify: cannot verify token


pascalgn:
We are using permission `user_impersonation` for our app registration when requesting an access token (`scope=api://123-our-client-id-456/user_impersonation`) and we get the following error:
```
authentication failed due to: In Strategy.prototype.jwtVerify: cannot verify token
```

When adding `loggingNoPII: false` to our config, we also see the following error:
```
authentication failed due to: jwt audience is invalid
```

dikokob:
any luck?

pascalgn:
@dikokob The issue should be fixed with #433, but nobody seems to care about it so far

pkanher617:
We are working on reviewing the remaining PRs in this repo. We are currently in the design phase of a new version of this library, which can be tracked [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/projects/9). The timeline for the new library will be uploaded to our [MSAL JS repo](https://github.com/AzureAD/microsoft-authentication-library-for-js).

PR #433, opened 2019-07-23
Add api://clientId to default audience array

pascalgn:
I want to preface this by saying I'm not at all an expert on Azure and/or Azure AD, but I think I've figured out a bug in passport-azure-ad.

What we did was basically follow this guide: https://docs.microsoft.com/en-gb/azure/active-directory/develop/v2-permissions-and-consent which talks about _the Application ID URI_. So let's say our App ID is `123-456-789`, then we requested the token with `scope=offline_access api://123-456-789/user_impersonation`. _Edit:_ After thinking about this some more, the audience is just given in the JWT as `aud`, so it's probably unrelated to the requested scope and just the App's URI.

This is the URI schema (`api://client-id`) that is also displayed in the Portal:

Screenshot 2019-07-23 at 20 33 41

But when using `passport-azure-ad`, we get the errors as described in #432 

So while debugging, I found this:

Screenshot 2019-07-23 at 20 03 19

That's why I think, when no explicit audiences are configured for this library, the version with `api://` should also be added.

### DIFF
--- a/lib/bearerstrategy.js
+++ b/lib/bearerstrategy.js
@@ -217,7 +217,7 @@ function Strategy(options, verifyFn) {
   if (options.audience && typeof options.audience === 'string')
     options.audience = [options.audience];
   else if (!options.audience || !Array.isArray(options.audience) || options.length === 0)
-    options.audience = [options.clientID, 'spn:' + options.clientID];
+    options.audience = [options.clientID, 'spn:' + options.clientID, 'api://' + options.clientID];
 
   // default value of isB2C is false
   if (options.isB2C !== true)


m4ss1m0g:
@pascalgn You are absolutely right ! After two hours trying to solve the problem, putting `api://clientID` on `audience` field solve the problem.  Unbelievable that the PR is not merged and/or this information is not  on README. Thanks.

m4ss1m0g:
> @pascalgn You are absolutely right ! After two hours trying to solve the problem, putting `api://clientID` on `audience` field solve the problem. Unbelievable that the PR is not merged and/or this information is not on README. Thanks.

After digging deep into the problem I found that the `api:` prefix is necessary when validating `id_token` NOT the `access_token`. If you enable debug you can see the validated token on console, when I read the token on (jwt.io)[http://jwt.io] I discovered the id_token.

jasonnutter:

Issue #375, opened 2018-04-12
Malformed JSON web token causes crash


Daniel-Hope:
version: 3.0.12
Hi,
I am using the Bearer Strategy to check auth tokens fetched from AAD.
The library works well and handles the situation appropriately when there is a valid or expired token sent in the authorization header.
Next I tried taking a valid token, deleting a chunk out of the middle of it, then sending that to my node server. This resulted in the jwt JSON parser throwing and error causing the node instance to crash.

I tried surrounding the `passport.authenticate` with a try catch and it did not stop it from crashing.

```
const express = require('express');
const router = express.Router();
const logger = require('winston');
const passport = require('passport');
const BearerStrategy = require('passport-azure-ad').BearerStrategy;

const AAD_AUTH_CONFIG = {
  identityMetadata: 'https://login.microsofto... etc.',
  clientID: '<app guid>'
};

passport.use(new BearerStrategy(AAD_AUTH_CONFIG, function(token, done) {
  return done(null, token);
}));

router.use(passport.initialize());

router.get('/path', (request, response) => {
  try {
    logger.info("checkAuthToken");
    passport.authenticate('oauth-bearer', function(err, user, info) {
      logger.info("inside callback");
      logger.info(err);
      logger.info(user);
      logger.info(info);
      response.send("Some Data");
    })(request, response);
  } catch(err) {
    response.send("Runtime Error");
  }
});
```

This is what the logs look like when I send it a request.
```
2018-04-12 14:06:30 PID:4161 INFO checkAuthToken
undefined:1
{"aud":"<app guid>","iss":"<valid value>","iat"3�$F�V ... garbled json
                                                                                                                         ^

SyntaxError: Unexpected number in JSON at position 121
    at JSON.parse (<anonymous>)
    at Object.jwsDecode [as decode] (/node_modules/jws/lib/verify-stream.js:71:20)
    at Strategy.jwtVerifyFunc [as jwtVerify] (/node_modules/passport-azure-ad/lib/bearerstrategy.js:290:23)
    at async.waterfall (/node_modules/passport-azure-ad/lib/bearerstrategy.js:568:19)
    at fn (/node_modules/passport-azure-ad/node_modules/async/lib/async.js:746:34)
    at /node_modules/passport-azure-ad/node_modules/async/lib/async.js:1213:16
    at /node_modules/passport-azure-ad/node_modules/async/lib/async.js:166:37
[...]

PR #436, opened 2019-08-06
Fix for 357, malformed json bug

MattheuM:
https://github.com/AzureAD/passport-azure-ad/issues/375

### DIFF
--- a/lib/bearerstrategy.js
+++ b/lib/bearerstrategy.js
@@ -287,7 +287,7 @@ util.inherits(Strategy, passport.Strategy);
 Strategy.prototype.jwtVerify = function jwtVerifyFunc(req, token, metadata, optionsToValidate, done) {
   const self = this;
 
-  let decoded;
+  let decoded = null;
   try {
     decoded = jws.decode(token);
   } catch (error) {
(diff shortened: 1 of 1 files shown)


herrmanntom:
Any change of this being merged? I have exactly the same problem. Its basically a DOS, as the entire node process crashes. And it cannot be handled by the code using this library.

lib/bearerstrategy.js
```diff
@@ -287,7 +287,13 @@ util.inherits(Strategy, passport.Strategy);
 Strategy.prototype.jwtVerify = function jwtVerifyFunc(req, token, metadata, optionsToValidate, done) {
   const self = this;
 
-  const decoded = jws.decode(token);
+  let decoded;
```
jasonnutter: To be caught in the `if` below, should this be instantiated to `null`? Or have the line 299 be updated to `if (!decoded)`

```suggestion
  let decoded = null;
```

jasonnutter:
This is available `passport-azure-ad@4.3.0-beta.0`. Please try it out and let me know if there are any problems, thanks!

PR #440, opened 2019-08-23
Limit number of packaged files

pkuczynski:
Does not seem like any of those are needed to be distributed? :)

I was also wondering about adding *.md here, as nobody reads them from unpacked package, but rather on the npm or github site...

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,6 @@
-# .npmignore file
+.idea/
+.vscode/
 test/
+.eslint*
+.travis.yml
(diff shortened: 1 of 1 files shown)


pkuczynski:
Cool, thanks! :)

jasonnutter:
@pkuczynski Have you signed this? https://cla.opensource.microsoft.com/AzureAD/passport-azure-ad?pullRequest=440

pkuczynski:
Just did...

PR #446, opened 2019-10-15
Added id and provider to makeProfileObject to comply with passport.js…

### LINKED ISSUE #445: Profile returned does not have 'provider' field populated
Hi,

According to the PassportJS documentation, a strategy should return a user profile with a 'provider' field populated(source)[http://www.passportjs.org/docs/profile/], in this case 'azure-ad' for example.

qbodart:
… documentation

fixes #445


jasonnutter:

PR #450, opened 2019-11-08
add support for samesite in passport js

DarylThayil:
To avoid breaking consumers who are using an http framework where this is not supported, I have added an opt in `cookieSameSite` option to OIDC config.
This will be a minor release.

### DIFF
--- a/lib/aadutils.js
+++ b/lib/aadutils.js
@@ -29,7 +29,7 @@ const util = require('util');
 exports.getLibraryProduct = () => { return 'passport-azure-ad' };
 exports.getLibraryVersionParameterName = () => { return "x-client-Ver" };
 exports.getLibraryProductParameterName = () => { return 'x-client-SKU' };
-exports.getLibraryVersion = () => { 
+exports.getLibraryVersion = () => {
   return '4.0.0';
 };
 
@@ -169,7 +169,7 @@ exports.rsaPublicKeyPem = (modulusB64, exponentB64) => {
 exports.checkHashValueRS256 = (content, hashProvided) => {
   if (!content)
     return false;
-  
+
   // step 1. hash the content
   var digest = crypto.createHash('sha256').update(content, 'ascii').digest();
 
@@ -256,7 +256,7 @@ exports.concatUrl = (url, rest) => {
   if (typeof rest === 'string' || rest instanceof String) {
     rest = [rest];
   }
-  
+
   if (!url) {
     return `?${rest.join('&')}`;
   }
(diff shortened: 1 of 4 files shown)


lib/aadutils.js
```diff
@@ -265,3 +265,62 @@ exports.concatUrl = (url, rest) => {
...
+
+// This is a list maintained by the AAD server team, will need to keep an eye on this
+// as things change.  Not ideal, but it is what it is
+//n general, a change like this (adding a new cookie attribute)
+// should be backward compatible as RFC 6265 specifies that browsers should
+// ignore unknown cookie attributes. However, for the specific case of the
+// SameSite attribute, some browsers incorrectly implemented the attribute or
+//  implemented an earlier draft which had contradictory behavior.
+// For these browsers which attempted to support SameSite,
+// but have bugs in their support, we want to omit the SameSite=None attribute.
```
hpsin: Add a link here to the Chromium team guidance so that people can get check it against the official guidance? https://www.chromium.org/updates/same-site/incompatible-clients

lib/cookieContentHandler.js
```diff
@@ -46,6 +47,11 @@ function CookieContentHandler(maxAmount, maxAge, cookieEncryptionKeys, domain) {
   if (!cookieEncryptionKeys || !Array.isArray(cookieEncryptionKeys) || cookieEncryptionKeys.length === 0)
     throw new Error('CookieContentHandler: cookieEncryptionKeys must be a non-emptry array');
 
+  if (typeof cookieSameSite !== 'boolean') {
+    throw new Error('CookieContentHandler: cookieSameSite must be a boolean');
+  }
```
jasonnutter: Instead of throwing, you can just cast whatever `cookieSameSite` is to its boolean equivalent: 

```js
this.cookieSameSite = !!cookieSameSite;
```

DarylThayil: yeah I could but my concern is that they pass in a string 'none' 'lax' 'strict'

lib/cookieContentHandler.js
```diff
@@ -25,16 +25,17 @@
...
  * @maxAmount          - the max amount of {state: x, nonce: x, policy: x} tuples you want to save in cookie
  * @maxAge            - when a tuple in session expires in seconds
- * @cookieEncryptionKeys  
+ * @cookieEncryptionKeys
  *                    - keys used to encrypt and decrypt cookie
  * @domain            - sets the cookie's domain
- */ 
-function CookieContentHandler(maxAmount, maxAge, cookieEncryptionKeys, domain) {
+ */
+function CookieContentHandler(maxAmount, maxAge, cookieEncryptionKeys, domain, cookieSameSite) {
```
hpsin: So is this off by default? I think it should be on, with option to disable.

DarylThayil: The concern was that it would need to be a major release if it was on by default. 
Not all http frameworks support the samesite option, so if i a customer was using express 3 for instance, this would throw.
They may be using other frameworks as well such as koa, hapi, restify along with passport.

Making this opt in means that when they go to update their app for samesite, this will be an option they need to confirm works with their dependency chain.

lib/aadutils.js
```diff
@@ -265,3 +265,62 @@ exports.concatUrl = (url, rest) => {
...
+      // - Chrome on iOS 12 for iPhone, iPod Touch, iPad
+      // All of which are broken by SameSite=None, because they use the iOS networking stack
+      if (userAgent.includes("CPU iPhone OS 12") || userAgent.includes("iPad; CPU OS 12"))
+      {
+          return true;
+      }
+
+      // Cover Mac OS X based browsers that use the Mac OS networking stack. This includes:
+      // - Safari on Mac OS X
+      // - Internal browser on Mac OS X
```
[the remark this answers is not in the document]
DarylThayil: copy pasta, not sure i caught anything :D

lib/aadutils.js
DarylThayil: will add

lib/aadutils.js
```diff
@@ -265,3 +265,62 @@ exports.concatUrl = (url, rest) => {
   return rest ? url.concat(hasParam ? '&' : '?').concat(rest.join('&')) : url;
 };
 
+
+// This is a list maintained by the AAD server team, will need to keep an eye on this
+// as things change.  Not ideal, but it is what it is
+//n general, a change like this (adding a new cookie attribute)
```
hpsin: **I**n general.  Typo :)

lib/cookieContentHandler.js
hpsin: Thanks, I appreciate the explanation.  We'll let developers know that this is a required action on their part in the notes.

lib/cookieContentHandler.js
```diff
@@ -133,6 +139,10 @@ CookieContentHandler.prototype.add = function(req, res, tupleToAdd) {
     options.domain = this.domain;
   }
 
+  if (this.cookieSameSite && !sameSiteNotAllowed(req.get('User-Agent'))) {
+    options.sameSite = 'none';
```
hpsin: Secure as well? We need to add secure to fully comply with Chrome changes.

lib/cookieContentHandler.js
jasonnutter: Maybe rename this to `enableSameSitCookie` to help make it clear that it should be a boolean?

jasonnutter: It actually looks like you are already casting to a boolean below, so this condition will never evaluate to false.

DarylThayil: so two things:
 1. for the name of the value, I chose the current name based on the other options for consistency  such as `cookieDomain` which is a similar option for setting cookie.  Also i left this sort of vague on purpose.  The express cookie for `sameSite` will take a `true` `false` OR `'none'` `'lax'` or `'strict'` .  Even though my current option only supports sameSite none  I wanted to leave the option in the future to open it up to one of these other values if requested by customers

2. Even though this is cast a level up,  I did want to make sure cookiecontenthandler enforced its contract in the scenario future consumers are not casting

lib/aadutils.js
DarylThayil: updated

lib/aadutils.js
DarylThayil: updated

lib/cookieContentHandler.js
DarylThayil: I will update readme as a part of the version cut to make sure this is all explained well

lib/cookieContentHandler.js
DarylThayil: added

PR #465, opened 2020-01-13
Strip nonbreaking spaces and fix a few lint errors

### LINKED ISSUE #464: sameSiteNotAllowed fails due to copy-and-paste artifacts
Version 4.2.1 is affected; I don't believe any others are based on git blame. The earliest commit that introduced the bug is cf04a064af8404970088ca5b236d6b3725d07b5b, I believe.

Run a script like this after installing `passport-azure-ad@latest` via `npm`:

```
const assert = require('assert').strict;
const { sameSiteNotAllowed } = require('./node_modules/passport-azure-ad/lib/aadutils.js');

const userAgent = "Mozilla/5.0 (iPhone; CPU iPhone OS 12_4_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/21.0 Mobile/15E148 Safari/605.1.15";
assert.equal(
  sameSiteNotAllowed(userAgent),
  true
);
```

This assertion will fail. It took me a while to figure out why, because when I rewrote it essentially verbatim, it worked fine.

To verify via the web without any code, look at the compiled source of this web page: https://github.com/AzureAD/passport-azure-ad/blob/dev/lib/aadutils.js

Observe the string that would normally match the user agent above is actually full of `&nbsp;` entities:
`"CPU&nbsp;iPhone&nbsp;OS&nbsp;12"`

Here's a picture of a hex dump highlighting the offending characters at the top of that function:

image

This failure causes cookie/session (and therefore state/nonce) problems for the targeted iOS 12, etc. web clients that are trying to use passport-azure-ad sessions. A workaround is to monkey-patch the function by taking advantage of `require()` caching:

```
const aadutils = require('./node_modules/passport-azure-ad/lib/aadutils.js');
function sameSiteNotAllowed (userAgent) {
  if (userAgent.includes("CPU iPhone OS 12") || userAgent.includes("iPad; CPU OS 12")) {
    return true;
  }
  if (userAgent.includes("Macintosh; Intel Mac OS X 10_14") && !userAgent.includes("Chrome/") && !userAgent.includes("Chromium")) {
    return true;
  }
  if (userAgent.includes("Chrome/5") || userAgent.includes("Chrome/6")) {
    return true;
  }
  if (userAgent.includes("UnrealEngine") && !userAgent.includes("Chrome")) {
    return true;
  }
  if (userAgent.includes("UCBrowser/12") || userAgent.includes("UCBrowser/11")) {
    return true;
  }
  return false;
}
aadutils.sameSiteNotAllowed = sameSiteNotAllowed;
```

bburhans:
Fixes #464 based on a cursory test of a single sample userAgent:

```
const util = require('util');
const assert = require('assert').strict;

const { sameSiteNotAllowed } = require('./node_modules/passport-azure-ad/lib/aadutils.js');

const userAgent = "Mozilla/5.0 (iPhone; CPU iPhone OS 12_4_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/21.0 Mobile/15E148 Safari/605.1.15";
try {
  assert.equal(
    sameSiteNotAllowed(userAgent),
    true
  );
  console.log(`${util.inspect(sameSiteNotAllowed)} returned true`);
} catch (err) {
  console.log(err);
}

const fixed = require('./aadutils.js').sameSiteNotAllowed;

try {
  assert.equal(
    fixed(userAgent),
    true
  );
  console.log(`${util.inspect(fixed)} returned true`);
} catch (err) {
  console.log(err);
}
```

### DIFF
--- a/lib/aadutils.js
+++ b/lib/aadutils.js
@@ -276,52 +276,46 @@ exports.concatUrl = (url, rest) => {
 // For these browsers which attempted to support SameSite,
 // but have bugs in their support, we want to omit the SameSite=None attribute.
 // See Chromium official guidance here:  https://www.chromium.org/updates/same-site/incompatible-clients
-exports.sameSiteNotAllowed = (userAgent) => {
-      // Cover all iOS based browsers here. This includes:
-      // - Safari on iOS 12 for iPhone, iPod Touch, iPad
-      // - WkWebview on iOS 12 for iPhone, iPod Touch, iPad
-      // - Chrome on iOS 12 for iPhone, iPod Touch, iPad
-      // All of which are broken by SameSite=None, because they use the iOS networking stack
-      if (userAgent.includes("CPU iPhone OS 12") || userAgent.includes("iPad; CPU OS 12"))
-      {
-          return true;
-      }
+exports.sameSiteNotAllowed = userAgent => {
+  // Cover all iOS based browsers here. This includes:
+  // - Safari on iOS 12 for iPhone, iPod Touch, iPad
+  // - WkWebview on iOS 12 for iPhone, iPod Touch, iPad
+  // - Chrome on iOS 12 for iPhone, iPod Touch, iPad
+  // All of which are broken by SameSite=None, because they use the iOS networking stack
+  if (userAgent.includes("CPU iPhone OS 12") || userAgent.includes("iPad; CPU OS 12")) {
+    return true;
+  }
 
-      // Cover Mac OS X based browsers that use the Mac OS networking stack. This includes:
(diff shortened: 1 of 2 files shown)


bburhans:
Turns out a test had the same bug. We should probably try to avoid copying and pasting when creating new tests, although I am guilty of it sometimes when cloning one test to another within the same file for good coverage.

jasonnutter:
This is available `passport-azure-ad@4.3.0-beta.0`. Please try it out and let me know if there are any problems, thanks!

PR #467, opened 2020-01-30
Version that supports SAML

bpcastilho:
Pushing information to Read.Me about the version that supports SAML and WSFED, for customer awareness.


jennyf19 (approved):
cool. thanks! :shipit:

PR #470, opened 2020-02-03
pin version of cache-manager

### LINKED ISSUE #469: params.metadata.generateOidcPEM is not a function
Without any code change, I started getting this error all of a sudden.

authentication failed due to: In _validateResponse: failed to generate PEM key due to: params.metadata.generateOidcPEM is not a function

I debugged and traced this to a newer version of cache-manager (>= 2.10.2) which has a change that breaks oidcstrategy.js. Specifically, the memorystore.js in the cache-manager does not mutate the object and instead does a deep clone thereby losing members of the prototype.

So, in this line 829 on oidcstrategy.js on function setOptions, it loses the prototype members leading to the error mentioned above.

params.metadata = metadata;

A simple Object.assign() in oidcstrategy.js must fix the issue.

I temporarily fixed by adding the cache-manager v 2.10.0 to my project.

paul-leathwood:
The dependency `cache-manager` changed from using `lodash.clonedeep` in `2.10.2` to `deepmerge` in `2.11.0`, this appears to have introduced some breaking changes around the `generateOidcPEM` method.

The simplest fix for this is to pin the version of `cache-manager` to `2.10.2`

This resolves #469

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "async": "^1.5.2",
     "base64url": "^3.0.0",
     "bunyan": "^1.8.0",
-    "cache-manager": "^2.0.0",
+    "cache-manager": "2.10.2",
     "jws": "^3.1.3",
     "jwk-to-pem": "^1.2.6",
     "lodash": "^4.11.2",


jennyf19 (approved):
:shipit:

jennyf19:
cc: @jmprieur

jmprieur:
LGTM.

ryanbecker:
This should be resolved in version `2.11.1` of `cache-manager`.

PR #472, opened 2020-02-05
fixing the SO tag here.

jennyf19:
MSAL was not really a product when passport.js was created, so made sense to have ADAL at that point as the SO tag. but we now should be using MSAL. Updating here so customers get answers faster.


jmprieur (approved):
Thanks @jennyf19 
LGTM

PR #474, opened 2020-02-06
Fix #473 - crash if bearer token is blank

### LINKED ISSUE #473: Crashes when Bearer token is blank in BearerStrategy
Version: 4.2.1

If the Authorization header is formed as: `Authorization: Bearer  `
(that is, just a blankspace after "Bearer") the token will be an empty string and will crash the server in the _authenticate_ method.

The issue is here: https://github.com/AzureAD/passport-azure-ad/blob/48d44bb978f16c9e25e8bdeac3efe874547fca27/lib/bearerstrategy.js#L529

It should return: 
```javascript
return self.failWithLog('In Strategy.prototype.authenticate: missing access_token in the header');
```

jthorel:
Fixes #473

### DIFF
--- a/lib/bearerstrategy.js
+++ b/lib/bearerstrategy.js
@@ -526,7 +526,7 @@ Strategy.prototype.authenticate = function authenticateStrategy(req, options) {
                 log.info(`In Strategy.prototype.authenticate: received access_token from request header: ${token}`);
             }              
             else
-              self.failWithLog('In Strategy.prototype.authenticate: missing access_token in the header');
+              return self.failWithLog('In Strategy.prototype.authenticate: missing access_token in the header');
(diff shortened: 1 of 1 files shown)


jennyf19:
@DarylThayil @jasonnutter can this one be merged as well?

jasonnutter:
@jthorel Have you signed this? https://cla.opensource.microsoft.com/AzureAD/passport-azure-ad?pullRequest=474

jthorel:
@jasonnutter Yes, just did!

jasonnutter:
@jthorel This is available `passport-azure-ad@4.3.0-beta.0`. Please try it out and let me know if there are any problems, thanks!

PR #531, opened 2021-01-14
Added new responseMode, responseMode=fragment

lflescano:
This option is missing so i added fragment to be a responseMode option


jasonnutter:
Response mode as fragment is only supported by our client side libraries, and will not work for server side libraries. Closing.

PR #544, opened 2021-05-17
Add option to force verify callback signature

0x0ece:
There are situation where the signature of the verify callback isn't correctly auto-detected. This PR adds an option to force the desired signature.

One relevant example is the NestJS framework, see [this issue](https://github.com/nestjs/passport/issues/53) caused by [this code](https://github.com/nestjs/passport/blob/master/lib/passport/passport.strategy.ts#L17).


jcarlin:
Any word on priority of this original issue and this PR? We cannot use passport-azure-ad with nest.js without this fix.

jasonnutter:

